PR #56, opened 2024-05-20
Creates logging folder and log file when docker image is built for npda

dc2007git:
Fixes bug where logger config wasn't functioning without an npda.log file.

Adds a startup script to create a logs folder and an npda.log file if they don't already exist, to alleviate this error for any users running the code locally.


dc2007git:
Closing after quick chat with @mbarton and a better solution proposed as in #57

PR #76, opened 2024-05-30
Adds autosave htmx functionality to Visit update/create form

dc2007git:
Adds htmx functionality to visit form to autosave when the user enters information into a field.

@eatyourpeas this is my first time integrating htmx into an application, could i just get you to quickly glance at what I've done and make sure you're happy with it? Cheers!

### DIFF
--- a/project/npda/templates/npda/visit_form.html
+++ b/project/npda/templates/npda/visit_form.html
@@ -5,7 +5,8 @@
 <div class="bg-rcpch_light_blue py-8">
   <div class="w-full max-w-3xl mx-auto px-2 py-4 m-2 shadow-md bg-white font-montserrat">
     <strong>{{title}}</strong>
-    <form id="update-form" method="post" {% if form_method == "create" %} action="{% url 'visit-create' patient_id %}" {% else %} action="{% url 'visit-update' patient_id=patient_id pk=visit.pk %}" {% endif %}>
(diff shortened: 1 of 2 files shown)


dc2007git:
Setting to draft as bug with visit form submission discovered, likely htmx issue

dc2007git:
Update - after discussion with @eatyourpeas we feel that the addition of HTMX to the form may produce more problems than solutions it would provide.

The initial rationale with htmx was that it can autosave the form, and so this would be useful especially with longer forms, much like the Audit form in Epilepsy12. However, the NPDA Visit form is much shorter and much faster to fill in, especially with the addition of buttons like 'Today' and 'Visit Date'. 

One reason why we wouldn't use htmx is because we would have to differentiate in the form whether we are creating or updating. If we use the form to create a new visit, then every time any field on the form is changed, a new 'create' instance is activated, which means that by the end of the form you will have upwards of 20 visits when you only meant to enter one.

Once a form has been opened with type create (ie you press 'Add new Visit'), and a field has been entered into, htmx will save a Visit. The next step would be to 'update' this form, but is difficult to seek for a visit.id that corresponds to the visit that has just been entered. One solution would be to simply pull the most recently created visit ID, but there is of course the chance that another user somewhere else will have just created a new Visit for a different patient, leading to editing of the wrong patient's visit.

To that end, we feel that at this point we do not need HTMX in the form. Perhaps at a later point if we feel it would benefit us then we can include it, but for the moment it is not worth the overhead.

eatyourpeas:
I agree. Let's park this for the moment.

PR #80, opened 2024-05-31
Adds Today button to each date field

### LINKED ISSUE #60: Pre-loading date into date fields causes that category to appear as 'addressed' in the Patient Visits section
The new feature in #48 where we preload the date fields into the form as today is causing that category to be listed as 'addressed' in the patients visit section like so:

image

dc2007git:
Purpose of PR is to remove a preset value for each date field, which was previously today's date. However, this activated the category, meaning that it showed up as 'addressed' even if the user had not entered anything - the preset value was enough for it to show user activity.

Instead, a button with 'Today' next to each date field has been implemented with Hyperscript to fill the date field if the user wishes so, fixing the bug above and also removing the number of clicks to enter today's date.

Closes #60


eatyourpeas (approved):
this is great nice work.
Job for next is to add a second button that offers an option to populate all dates with the visit date

PR #81, opened 2024-05-31
Change design of Visit form to load errors into form

### LINKED ISSUE #77: Add flagging of errors to Visit form
[discussion with @eatyourpeas]

In the Visit form, before you click on a visit you can see which categories of the form have errors. However, once you enter the form, you can't see anymore which of those fields need correction.

We should have some way for the user to see the errors whilst they're in the form, potentially highlighting the area that is wrong in red?

dc2007git:
Before this PR, a user would have to go to a patient, and view all visits, in order to see which parts of the Visit form they have submitted have errors / aren't validated. These parts of the form, which we have named 'categories', would be highlighted in red if they had errors. However, when the user clicks on the form, they would have to memorise which parts of the form had errors as this information was not sustained in anyway on the next page by error messages or anything similar.

This PR introduces a colour-coordination to this problem. Any part of the form that has not had information filled into it will appear as dark blue, any part of the form where there are no errors (and the form is valid) will be pink, and any category part of the form where there are errors will show up as red.

This PR also incorporates a couple of minor bug fixes, such as removing rounded corners from the accordion in line with RCPCH styling.

Closes #77

### DIFF
--- a/project/npda/templates/npda/visit_form.html
+++ b/project/npda/templates/npda/visit_form.html
@@ -2,8 +2,8 @@
 {% load npda_tags %}
 {% csrf_token %}
 {% block content %}
-<div class="bg-rcpch_light_blue py-8">
-  <div class="w-full max-w-3xl mx-auto px-2 py-4 m-2 shadow-md bg-white font-montserrat">
+<div class="flex justify-center bg-white py-8">
+  <div class="w-full mx-96 px-2 py-4 m-2 shadow-md bg-white font-montserrat">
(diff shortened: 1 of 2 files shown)


eatyourpeas (approved):
This is great @dc2007git. I will merge this but still work in progress. In particular, if all the tiles are to be pink if complete, red if invalid and dark blue if unscored, we should roll that same style across all the fields in each tab.
Also the red and pink clash  quite a lot, so we need to think about ways to make it easier on the eye...

Issue #40, opened 2024-05-06
Less hacky way to skip OTP in development


eatyourpeas:
Although the workflow is largely the same as E12, it does not seem to work quite the same as NPDA.
Currently although the OTP field is populated with PASSED in development mode, it does not bypass the OTP work flow and verify the user.

anchit-chandran:
this is fixed (in a hack-y) by #82 - involves overriding the `post()` method of the `RCPCHLoginView` class.

Will keep issue open but backburner to indicate that a more 'proper' solution would be ideal, if any one has ideas

anchit-chandran:
Closing this as it's behind a `settings.DEBUG` flag

PR #82, opened 2024-06-01
Bypass 2FA Token in local development

anchit-chandran:
Got working with help of @eatyourpeas thanks!

### Overview

When in local development (`settings.DEBUG=True`), bypass the 2FA token step on login. 

This improves the developer experience.

### Code changes
- Update various URLs in both project and `npda` app to only use 2FA / custom auth urls
- Adds `LoginAndOTPRequired` and `login_and_otp_required` mixin&decorator
- Specifically overrides the `.post()` method of the `RCPCHLoginView` class to skip past the normal auth workflow if in debug mode

- nil

### Related Issues
- #40


eatyourpeas (approved):
Thank you Anchit for this nice piece of work. Looks all good.
One thing I have changed - would like to know whether you agree - is that I have removed the requirement to be both DEBUG and superuser to skip 2FA. My rationale is that 2FA is for production only, when DEBUG will be False, and I have found otherwise that if we are testing with different user accounts, it throws a 403.

PR #95, opened 2024-06-04
Eatyourpeas/issue86

### LINKED ISSUE #86: RCPCH and superuser national access
Currently the PDU relating to the user is stored in session and used to scope which users and patients are seen down to the PDU.

Superusers and RCPCH team members need to be able to change the PDU/organisation in the session. One suggestion here is for some HTMX and a custom threeway button component with a select, as in E12. The idea would be that a user with national access can view individual organisations within a PDU, the whole PDU as an entity, or nationally (ie all patients/users). This means:
1. a new threeway button and select as a partial
2. new htmx post urls to fire on click, with a custom function in the header out of band to refresh the table.
3. logic in the views POST methods to update the session and return the updated partial to the template.
4. add permissions to these endpoints to ensure only the right users can access.

eatyourpeas:
### Overview
The patient list is now filtered against view preference which is stored in the NPDAUser model.
`view_preference` has 3 options: 0 (organisation), 1 (PDU) or 2  (National). RCPCH Audit team members or staff or superusers have national view access, everyone else can view patients across their PDU, which may contain more than one organisation. Within this, it is now possible to select between these organisations and filter patients to them. If logged in as as superuser or RCPCH user, it is possible to view all patients nationally, or filter across the PDUS and their child organisations.

This PR introduces HTMX that has a new 3-way button partial and select below it. The table itself has also be abstracted out into its own partial. On login a user's employer organisation and their view preference and whether they are an RCPCH audit team member or superuser are used to poll the RPCH NHS Organisations API and store their PDU and Organisation ODS Code in Session, as well as lists of sibling organisations (or all organisations and pz codes if a superuser) to populate the dropdown(s).
The click event posts to a new `post` method in the `patients` ListView which checks if htmx headers are present and pulls the hx-headers to identify which button was pressed, update the session from the API and return the view_preference partial with the updated state. This process also triggers a new custom trigger which leads to a new get request from the `patients_template`. The new `get` function in the patient listview requeries the Patient table with updated filters before returning the table partial.

In effect, this merges all the htmx callbacks to the same class, allowing all the permissions mixins/decorators to be used only once. The permissions still need updating and there is a separate issue for this.

### Code changes
`settings.py` - adds htmx_middleware to `INSTALLED_APPS`
`retrieve_pdu` - new function to retrieve a PDU and its child organisations
`mixins` - fix to allow all logins to bypass 2FA in DEBUG mode if authenticated
`npda_users` - additions to RCPCHLogin class to hit RCPCH NHS Organisations API for organisation and PDU data to store in session.
`views/patient` - new methods to `patients` ListView: a new get and post method to use htmx middleware to interrogate the headers, update the session and return the partials
`templates/partials` - a new folder containing two new partials (`view_preference` containing threeway toggle button and select for pdus or organisations) and `patients_table`). The `view_preference` partial and work flow will be reused for the NPDA User workflow later
`base.html` - new tailwind settings to match width of buttons to width of select in `view_preference` This is not great and will need reworking

This will need adding

### Related Issues
Closes #86

### Future items that flow from this and specific questions for code review prior to merge (if this feature is ok):

1. the same workflow needs to apply to the user table
2. the API calls can be a little slow - is there any optimisation either here or in the API to improve this?
3. is using session in this way an acceptable way to pass state around the application?
4. styling - the threeway toggle and select style is copied from E12 but does not quite work: feedback on better implementations of look and feel
5. styling - we need spinners within the buttons while the htmx requests are in flight. We have the same in E12 which works well and is a simple addition.


eatyourpeas (commented):
Review with @mbarton
Discussion:
1. it is fine to use session to store user details
2. the idea of having a top-level scope that can be changed from anywhere (view_preference) is good
3. HTMX is fine but possibly overkill for what we are trying to achieve. Arguably could store more in the url and trigger a page refresh than necessarily updating only small bits of the screen. Either way, plan for the moment to stick with what we have.

Issue #102, opened 2024-06-06
Pre-fill all login fields for local development


anchit-chandran:
Having to keep typing in fields / even pre-filling is tedious for local development!

Pre fill these fields, don't require captch if `settings.DEBUG` is True

mbarton:
Have we checked the django secret is set? If it is you shouldn't have to log in again on each restart which is the most annoying thing

anchit-chandran:
closing as done in PR

PR #103, opened 2024-06-06
Faster-login-local-dev

anchit-chandran:

### Overview
Filling login fields is tedious for local development. This PR simply pre-fills fields, as per .env file values. Have added these to the `example.env`:

```
# FOR LOCAL DEVELOPMENT
LOCAL_DEV_ADMIN_EMAIL="incubator@rcpch.ac.uk"
LOCAL_DEV_ADMIN_PASSWORD="BR#GbdPH5j#utYShr&%QUB$*BS#YbLA8w6QxN8H4"
```

NOTE: requires creating a superuser with the above values and also having these in your .env. Nothing breaks if you dont', but the pre-filled fields won't match the superuser details so not much use.

### Code changes
Primary changes in `npda_user_form.py`

image

- Updated docker-setup 

image

### Related Issues
#102 

### Mentions


envs/example.env
```diff
@@ -37,4 +37,8 @@ DEBUG="True" # Set DEBUG=True for Local dev and Development, not Staging or Live
 
 DJANGO_STARTUP_COMMAND="python manage.py runserver 0.0.0.0:8008" # for local development with auto-reload
 # DJANGO_STARTUP_COMMAND="gunicorn --bind=0.0.0.0:8000 --timeout 600 rcpch-audit-engine.wsgi" # for live deployment
-NPDA_SECRET_KEY=##########
\ No newline at end of file
+NPDA_SECRET_KEY=##########
+
+# FOR LOCAL DEVELOPMENT
+LOCAL_DEV_ADMIN_EMAIL="incubator@rcpch.ac.uk"
+LOCAL_DEV_ADMIN_PASSWORD="BR#GbdPH5j#utYShr&%QUB$*BS#YbLA8w6QxN8H4"
```
mbarton: Minor: let's make this an obviously insecure string like "devpassword" just so it doesn't look at a glance like a real password has been committed to the repo

eatyourpeas: I agree. I will make that change. Otherwise looks good to me. Also, in a separate commit elsewhere I have corrected the env in settings.py that pulls in the secret from .env also so should not have to login each autoreload

eatyourpeas (approved):
this all looks good to me.
I have replaced the password as per @mbarton. Autoreload should not force login either as have added the correct secret to settings.py in a separate commit elsewhere now in development branch

PR #116, opened 2024-06-11
audit-cohort-implementation

### LINKED ISSUE #97: csv_upload does not remove previous visits
Currently csv_upload overwrites existing Patient records, but does not remove the attached visits, leading to orphan records. 

To fix this:
On csv_upload, if a Patient instance is being updated, rather than created, the associated visits should be removed first

eatyourpeas:
### Overview
Far more commits here than necessary and that is because it went through multiple experiments before settling on this.

This introduces a new model class, AuditCohort, which has a many to many relationship with Patient. There is an instance per upload, though there can only be one valid record per audit year and cohort.
The cohorts are quarterly within each year. 

### Code changes
Please describe the changes you made here.

`settings.py` corrects the secret key to prevent multiple logins
`patient.py` refactor PatientListView to reflect filter against current audit year and cohort
`home.py: introduce a summary page on upload of csv that counts unique patients
`cohort_for_date`: new function for cohort calculation
`audit_cohorts`: new list view for AuditCohorts. Allows post request to delete instances from the table. This deletes associated patients and updates the most recent to be valid. Only the valid field in the table can be viewed in the patients list. In future it will hold historical submitted cohorts.

templates
new `audit_cohort_list` template and its table partial.
some updating and restyling to existing templates as well as updating fields to match new model

caddy/postgres
refactor names in docker compose to allow other rcpch dockerised instances to run in parallel

### Related Issues
closes #97


project/npda/general_functions/cohort_for_date.py
```diff
@@ -0,0 +1,25 @@
...
+    if date_instance < audit_start_date:
+        # The patient was audited in the previous year
+        days_remaining = (audit_start_date - date_instance).days
+    else:
+        # The patient was audited in the current year
+        audit_end_date = date(date.today().year + 1, 3, 31)
+        days_remaining = (audit_start_date - date_instance).days
+    total_days = (audit_end_date - audit_start_date).days
+    completed_days = total_days - days_remaining
+    if (days_remaining / completed_days) < 0.25:
```
dc2007git: Just sanity checking myself here. NPDA will only have 4 cohorts? Also how does the percentage ratio of how many days remaining / completed_days dictate which cohort the patient will be in? It looks like if there is less than 25% of the days left in the cohort, then they're part of cohort 1, and this increments up by 25% each time until cohort 4 gets assigned?

eatyourpeas: this will possibly need a test. Every audit year currently (which runs from 1st April to 31st March) has only one submission, though people often upload more than one submission as they make revisions to the original submission. The NPDA team asked that we keep a record of all submissions made for a given audit year.
They also told us that HQIP are requesting that the implement quarterly reporting for a given audit year, so this is why I have introduced the idea of a cohort. The `audit_year` therefore is the year of submission (April-March), the cohort is that same year divided into quarters. There is first a check on the date of the request, to see if it is before or after April.
Any CSV uploaded is allocated an `audit_year` and `cohort_number` based on the date of upload, though only one of these can be `submission_active` (the latest).

It is possible that this function does not do what it intended. There is already a rudimentary pytest suite, and this does not require a database to test, so possibly we could write some tests for this and amend the function as necessary, in true test-driven development style...

dc2007git: Gotcha - cohort is the quarter of the year submitted! A little bit of the E12 concept of cohort was leaking into my thought process but that makes sense. Perhaps it's worth a comment in the code explaining it like you have just to avoid any confusion for other devs working between e12 and npda. Happy to do this myself as I review the PR now.

The logic for the function looks good to me - but always happy to go for TDD as a sanity check! Only suggestion I have after a review is switching the numbers for cohort around - so if you have <25% days left, then you're in the 4th quarter and as such should be cohort 4, whereas at the moment for this case you would be assigned 1. Again happy to switch this around.

This will of course require another migration and the way we've been doing migration updates in this repo, looks like we're keen to keep the number of migrations low, nuking the current number in favour of having a single tidy initial migration. Is doing this just the case of deleting this current migration once changes have been made, then running makemigrations --> migrate @eatyourpeas ?

Thanks!

project/npda/models/auditcohort.py
```diff
@@ -0,0 +1,108 @@
...
+        Returns 4 if the patient has more than 75% of the audit year remaining
+        """
+        audit_start_date = date(self.audit_year, 4, 1)
+        if self.submission_date.date() < audit_start_date:
+            # The patient was audited in the previous year
+            days_remaining = (audit_start_date - self.submission_date.date()).days
+        else:
+            # The patient was audited in the current year
+            audit_end_date = date(self.audit_year + 1, 3, 31)
+            days_remaining = (audit_start_date - self.submission_date.date()).days
```
dc2007git: Only potential logic error I can see might be here, and on the same code above in the if statement (line 80). If we're calculating days remaining in this current cohort shouldn't we be using `audit_end_date` instead of `audit_start_date`?

project/npda/general_functions/cohort_for_date.py
eatyourpeas: Great thoughts thank you. Go ahead and change things you feel make sense. I don't feel strongly about migrations but generally while we are moving fast and breaking things I tend to start with a fresh DB and therefore like the idea of very few migrations. As things slow down then yes we should retain the changes.

dc2007git: Great thanks Simon - will do. Just to double check before I implement - we're splitting up the audit year into quarters and simply assigning patients a quarter (aka cohort) dependent on which quarter their submission date falls into?

And for completed days - are we calculating the number of days since the patient's submission (aka Today - submission_date) or the number of days into the audit year the patient is (submission_date - audit_start_date)?

project/npda/templates/partials/patient_table.html
```diff
@@ -101,6 +103,12 @@
                         {% endif %}
                         {{ patient.diagnosis_date }}
                     </td>
+                    <td class="px-6 py-4 text-center">
+                            {{patient.last_upload_date|date:"d/m/Y H:m s"}}
+                    </td>
+                    <td class="px-6 py-4 text-center">
```
dc2007git, line 109: Could be nice to have a brief description or tooltip defining the cohort number here

project/npda/views/patient.py
```diff
@@ -15,13 +17,13 @@
 # Third party imports
 from django_htmx.http import trigger_client_event
 
+from project.npda.general_functions.cohort_for_date import retrieve_cohort_for_date
```
dc2007git: This import has been added but not used anywhere

project/npda/views/patient.py
```diff
@@ -15,13 +17,13 @@
...
 from project.npda.general_functions.rcpch_nhs_organisations import (
     get_all_nhs_organisations,
```
dc2007git: Also this import is no longer used

project/npda/views/patient.py
```diff
@@ -45,37 +47,37 @@ def get_queryset(self):
...
         else:
             raise ValueError("Invalid view preference")
 
+        patient_queryset = Patient.objects.filter(
+            audit_cohorts__submission_active=True,
+        )
+        if filtered_patients is not None:
+            patient_queryset = patient_queryset.filter(filtered_patients)
+
+        AuditCohort = apps.get_model(app_label="npda", model_name="AuditCohort")
```
dc2007git: Does this instance of AuditCohort need to be defined here? I can't see it referenced anywhere else

project/npda/models/auditcohort.py
dc2007git: This has been actioned in my most recent commits to this PR. Will wait for confirmation from @eatyourpeas then happy to resolve this conversation.

project/npda/templates/partials/audit_upload_history.html
```diff
@@ -0,0 +1,66 @@
...
+                            <td>{{ audit.submission_date|date:"d/m/Y H:i s" }} secs</td>
+                            <td>{{audit.submission_by}}</td>
+                            <td>{{audit.audit_year}} ({{audit.cohort_number}})</td>
+                            <td>{{audit.patient_count}}</td>
+                            <td>{{audit.submission_active}}</td>
+                            <td class="flex flex-row">
+                                <form method="post" action="{% url 'audit_cohorts' %}" class="join">
+                                    {% csrf_token %}
+                                    <input type="hidden" name="audit_id" value="{{audit.pk}}">
+                                    <button name="submit-data" value="download-data" type="submit" class="join-item bg-rcpch_light_blue text-white font-semibold hover:text-white py-1 px-2 border border-rcpch_light_blue hover:bg-rcpch_strong_blue hover:border-rcpch_strong_blue btn-sm rounded-none">Download</button>
```
dc2007git, line 30: Just signposting that this button doesn't yet have functionality. Delete button below works fine though.

dc2007git (commented):
Happy with the rest of this PR, have tested it locally myself and aside from comments above everything seems to be in order. Nice job @eatyourpeas !

eatyourpeas:
Thanks @dc2007git that is great. Feel free to make any edits/changes as you see sensible and merge. You are right about the redundant button - needs implementing: perhaps warrants a separate issue.

project/npda/views/patient.py
eatyourpeas: if not used agree, do remove.

project/npda/templates/partials/audit_upload_history.html
eatyourpeas: do please raise a separate issue for this if you have time

project/npda/templates/partials/patient_table.html
eatyourpeas: I agree nice idea. Worth raising a separate issue. I have handwritten all our tooltips in css but actually would be better to refactor and replace all these with daisyui ones which are nicer. Again, a separate issue for this is a good idea

dc2007git (commented):
Merged after testing locally. Note @devs , will require a nuking of docker volumes

Issue #13, opened 2024-04-15
Permissions mixin


eatyourpeas:
Allied to #12 
As we are using class-based views, we can leverage Permissions mixins. These need implementing to constrain views so that:
1. only logged in users can view endpoints
2. users can only view/edit/delete (based on permission) children in their own PDU
3. rcpch audit team can view/edit/delete nationally

eatyourpeas:
The django LoginPermissionsMixin has been added to all views in PR #19 
A custom permissions mixin will be needed to constrain users access to their own organisations children, as well as what they can do within their organisation, based on their role

mbarton:
Implemented in #149

PR #129, opened 2024-06-15
Permissions integration

dc2007git:
### Overview

This PR will integrate user permissions into different views. Taking advantage of our class-based view system, we can utilise the `PermissionRequiredMixin` which enables us to load a view only once a user's specific permission (which is specified in that very view) is checked. Else, return (at the moment) 403.

n.b. if you wish to test these permissions, you must delete all containers, images AND volumes on your device and rebuild from there for them to be integrated.

### Code changes

`create_groups.py` - assign user permissions to each group that exists in NPDA. I have removed the old system of permissions (which we use in E12), where we have a set of permissions for 'viewing' all models, 'creating' all models etc., and then assigning each user group access. The approach implemented in this PR is more specific and explicit, allowing for simple permissions customisation in the future if we wish to do so, but at the expense of more lines of code.

`mixins.py` - slightly increase verbosity of logging when a user enters a view, showing their role and superuser status

`views/patient.py` - implement permissions checks

`views/visit.py` - implement permissions checks

`settings.py` - remove duplicated `ALLOWED_HOSTS`

`mixins.py` - add CheckPDUMixin for list views. 

Required - update permissions in docs to specify NPDAUser-related perms

### Related Issues

#13


dc2007git:
@eatyourpeas can I ask for your review on this? Specifically the `CheckPDUMixin` in mixins.py and its functionality - remember we were trying to stop bad actors from changing session variables to try and access PDUs that they shouldn't be able to.

Also, before we merge I'm just waiting for Amani to get back to me about the new user groups - Coordinator, Editor, and Reader. If she says so, then before we merge this PR we'll have to implement those groups instead of using Lead Clinician, Clinician etc. as we have. 

So just really requesting a general functionality review please!

eatyourpeas:
Great start @dc2007git  - so far two things:
1. Here I have created a user jane.admin@nhs.net with coordinator permissions at Addenbrookes. She is able to update her own record to make herself superuser/rcpch_audit_team_member and so on.
image


2. If I then update the url to the unique id for me I am able to edit a different record - I am registered at King's so this should not be possible.
image

That must mean that the mixin is letting me through. Have a look and we can try and troubleshoot together if no dice.

dc2007git:
Thanks for the review @eatyourpeas . Great points, thanks for picking up on them!

1. Users definitely should not be able to change their own records to make themselves superusers. I have recreated a user of type Coordinator, and have been able to reproduce this. Coordinators DO have the permission set to change_npdauser, which means they can technically access their own profile - and so possibly what we need is to only render the is_superuser / is_admin checkbox for current superusers. Is it also a good idea to stop those with the change_npdauser permission from changing their OWN information (ie they have to contact the rcpch team)?

2. For the Update view, the mixin isn't in place! So should be an easy case of just adding it, looking something like this: check if npdauser1 with pdu1 is either: in the same pdu as npdauser2 OR superuser OR rcpch audit team.

Let me know if you have any other thoughts :)

project/npda/templates/npda/npdauser_form.html
```diff
@@ -20,6 +20,12 @@
                 </select>
               {% elif field.field.widget|is_textinput %}
                 <input type="text" id="{{ field.id_for_label }}" name="{{ field.html_name }}" {% if field.value %} value="{{ field.value }}" {% endif %} class="input rcpch-input-text">
+              {% elif field.field.widget|is_checkbox and request.user.is_superuser and field.id_for_label == 'id_is_superuser' or field.id_for_label == 'id_is_staff' %}
```
dc2007git: @eatyourpeas completely see what you're doing here Simon and is exaxtly what the form needs. Doesn't look like it's working though - for my user of type Coordinator (who does not have is_staff or is_superuser ticked) they are able to see all 4 checkboxes and change them as they wish

project/npda/templates/npda/npdauser_form.html
```diff
@@ -20,6 +20,12 @@
...
+                {{ field }}
+              {% elif field.field.widget|is_checkbox and request.user.is_rcpch_staff and field.id_for_label == 'id_is_rcpch_staff' %}
+                {{ field }}
+              {% elif field.field.widget|is_checkbox and request.user.is_rcpch_audit_team_member and field.id_for_label == 'id_is_rcpch_audit_team_member' %}
+                {{ field }}
               {% else %}
                 {{ field }}
```
dc2007git, line 32: We need to remove this else statement as I believe this is what is causing the checkboxes to render for those people who it shouldn't render for. An elif statement above checks if a user is a superuser AND if the field.id_for_label is 'is_superuser', and if this resolves to False, then the else statement will load in the field anyway at the end.

We will just have to explicitly load the email field (as this is the only field not of type select, textinput or checkbox and thus would only be rendered by the problematic else statement unless explicitly called)

dc2007git:
@eatyourpeas another thought - we should perpetuate this idea of users with add/change_npdauser permissions only being able to add/change users to permissions equal to or lower than theirs. So a Coordinator should not be able to add an RCPCH Audit Team member, for example.

eatyourpeas (commented):
this is better great job. the labels for those fields still show and also the role drop down also still allows the admin user to upgrade themselves to rcpch user.
remember whatever we mask from the user in the UI does not hypothetically prevent them from passing directly to the API in postman or something. so what we are doing here is only protecting the user from making the wrong choices, rather than actually anything security - your mixin will do that: this works for me so that is great.

eatyourpeas:
Here is a screenshot from postman to illustrate what I mean. Here you can see I am posting data to a different user and correctly getting a 403. Yes you are right that we need to lock down people of escalating their own privileges. This will have to be done probably with the function, rather than the mixin.
image

dc2007git:
> Here is a screenshot from postman to illustrate what I mean. Here you can see I am posting data to a different user and correctly getting a 403. Yes you are right that we need to lock down people of escalating their own privileges. This will have to be done probably with the function, rather than the mixin.

Amazing thanks so much @eatyourpeas . Will be good to test things like the htmx through postman!

eatyourpeas:
hi @dc2007git is this meant to be closed?

mbarton:
yes closed in favour of https://github.com/rcpch/national-paediatric-diabetes-audit/pull/149 as @dc2007git had to fix a rebase issue

PR #131, opened 2024-06-17
Add errors into Visit form for error localisation

### LINKED ISSUE #123: UI change - error rendering in Visit form
[after dev team discussion]

- [x] 1. When a user is looking at the Visit List view for a patient, it is clear where the errors are, and hovering over the red tabs for the erroneous categories shows the error message. Persisting these error messages somewhere into the form would be very helpful for the user to know what is wrong with the current state of the visit.

- [x] 2. The errors do present themselves, but only on form submission. So if the user goes onto the form, the error messages are not present, and they would have to submit the erroneous form again for the errors to pop up.

- [x] 3. When the errors do pop up as described in point 2, they are not formatted well. Tidying them up to make them inline would be nicer.

- [x] 4. Additionally, it isn't [totally] obvious that there are tooltips over the errors which contain the exact error messages. It might be nice to have a 'Hover for Error message' text in the column header:

image

dc2007git:
### Before this PR

Errors would not load into the Visit form, and the user who is reviewing an entered visit would have to dart back and forth between the VisitsListView template and the Visit itself.

### After this PR

Errors now load into the form in the category where they are relevant. They do however still need formatting, as they are raw errors that are not incredibly user-friendly:

image

n.b. these are also the errors in the tooltips in the VisitsListView. Tidying them up (potentially in the template tag which finds and loads them into the template) would fix both the tooltips and form in this regard.

### DIFF
--- a/project/npda/templates/npda/visit_form.html
+++ b/project/npda/templates/npda/visit_form.html
@@ -51,35 +51,41 @@
                 outline outline-8 outline-rcpch_red
               {% endif %}"
             >
-            {% for field in form %}
+              {% for field in form %}
                 {% if field.field.category == field_category %}
-                <div class="flex flex-row my-2 mx-2">
-                  <div class="flex items-center justify-center md:w-1/3">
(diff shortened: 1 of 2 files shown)


eatyourpeas (approved):
looks good. i agree we need to think about styling

Issue #170, opened 2024-07-10
move tooltip positioning as covering npda user id in user table


anchit-chandran:
Annoyingly the tooltip covers the link to click into a user via their ID. 

![Image](https://github.com/rcpch/national-paediatric-diabetes-audit/assets/68898006/b74dbfc0-7a7c-4f00-a3a1-5e94f16a0d2d)

Fixed in the `perm-tests` branch:

![Image](https://github.com/rcpch/national-paediatric-diabetes-audit/assets/68898006/77cf2989-db69-4ff9-b674-79ff0b18e7e5)

mbarton:
Would https://github.com/rcpch/national-paediatric-diabetes-audit/issues/150 help here?

mbarton:
Not sure what fixed this but the tooltip is now fully to the right

PR #172, opened 2024-07-10
Adding tests for NPDAUserListView, cleaning up session access

anchit-chandran:

### Overview
This implements further foundations for testing the application, including 2 tests which ensure the `.get_queryset()` method for the NPDAUserList view returns the correct users for the given requesting user, based on their registered PDUs. To get this working, multiple updates to the test setup were made e.g. NPDAUserFactory.

More, in setting up these tests, realised the session object creation could be standardised: thus, created a global function `create_session_object_from_organisation_employer` which returns a dictionary that can be used to update the session, as seen in the RCPCHLoginView. This also means changes to session attributes can be made in one place and ensures the same method to create a session is tested.

Added the `Session` model to be viewed in the Django Admin. This has helped explore better implementation / use of the session.

### Code changes
Other misc changes include:
- Fix tooltip issue [#170](https://github.com/rcpch/national-paediatric-diabetes-audit/issues/170)

 ** Updating User perms table in docs**

Found it easier to separate into different tables, per role:

### (OLD)
image

#### (NEW)
image

#### Related Issues
[#170](https://github.com/rcpch/national-paediatric-diabetes-audit/issues/170)

### Mentions
@mbarton


anchit-chandran:
Just noticed all of these commits are `unverified` (joys of setting up git on WSL2). will fix

anchit-chandran:
GPG Verification with github set up now!

project/npda/templates/partials/npda_user_table.html
```diff
@@ -1,7 +1,7 @@
 {% load npda_tags %}
 {% if npdauser_list %}
-    <table class="table-auto w-full text-sm text-left rtl:text-right text-gray-500 text-gray-400 mb-5 font-montserrat">
-        <thead class="text-xs text-gray-700 uppercase bg-gray-50 bg-rcpch_dark_blue text-white">
+<table class="table-auto w-full text-sm text-left rtl:text-right text-gray-500 text-gray-400 mb-5 font-montserrat">
```
mbarton, line 3: has anything changed in this table or just a reformatting?

project/npda/views/npda_users.py
```diff
@@ -563,21 +543,17 @@ def done(self, form_list, **kwargs):
...
-            self.request.session["ods_code"] = current_user_ods_code
-
-        if "pz_code" not in self.request.session:
-            self.request.session["pz_code"] = current_user_pz_code
 
-        if "sibling_organisations" not in self.request.session:
-            sibling_organisations = get_single_pdu_from_ods_code(current_user_ods_code)
-            # store the results in session
-            self.request.session["sibling_organisations"] = sibling_organisations
+        request.session.update(new_session_object)
```
mbarton: I'd just refer to `self.request` directly if we can as that matches elsewhere in the code

project/npda/templates/partials/npda_user_table.html
anchit-chandran: unsure why it's not diffing properly but essentially just changed the position of the tooltip to be to the right soit doesn't cover the user id link

project/npda/tests/permissions_tests/test_npda_user_model_actions.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Extract the queryset from the context
+        view = NPDAUserListView()
+        view.request = response.wsgi_request
+        view.request.user = test_user  # Explicitly set the user in the request
+        queryset = view.get_queryset()
+
+        # Ensure the queryset does not contain users from a PDU they are not part of
+        assert (
+            queryset.exclude(organisation_employers__pz_code=ALDER_HEY_PZ_CODE).count()
```
mbarton, line 67: Minor, feel free to ignore.

I had a bit of trouble scanning the intent here, maybe spinning it around to add an include for the GOSH PZ code and checking the count is still 3 is clearer?

anchit-chandran: Was in process switching this but I think this is more robust, in terms of if we update the seeded test database / ODS codes change etc.

This currently checks if users from ANY other organisation other than Alder Hey are present.

Switching to checking for a specific one could lead to tedious / even missed bugs

mbarton (approved):
Thanks, looks good!

PR #179, opened 2024-07-11
Fix login

mbarton:
Broke in #172 but only for the non DEBUG path login so we didn't notice until it was deployed to staging.

```
UnboundLocalError: cannot access local variable 'user' where it is not associated with a value
```


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 264 hours, 44 minutes and 27 seconds ago) Please check your changes!

PR #184, opened 2024-07-11
Fix PDU filter in mixin

mbarton:
`values_list` by default returns a tuple, even if there's only one field in each entry. As such the check lower down didn't work, as it was looking for the PDU string in a list of tuples.

The fix is to pass flat=True which makes it return a list of values not tuples.

Before this fix a standard user (whatever role but not a superuser or RCPCH team) would get a full-screen 403 accessing either the list of patients or users.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 260 hours, 5 minutes and 19 seconds ago) Please check your changes!

PR #188, opened 2024-07-15
Filter PDU switcher options

### LINKED ISSUE #185: Standard users can select PDUs they don't belong to
1) Create a standard user in one PDU
2) Upload patient data to another (either as a superuser or standard user)
3) Log in as the first user

**Expected**: you can only select PDUs with organisations that employ you
**Actual**: you can select any PDU you like

You can't actually see the patient data from the other PDU as the reload triggered returns a 403. The UI gets stuck with the selector showing the new PDU but showing the patient data from the old one.

Note the tests in #172 only check that the data is filtered correctly. We will need additional ones to check you can only select PDUs you have access to.

We should also check you cannot manually craft a request to set your view_preference to all, even though the option is not available from the UI. The same for selecting a PDU just by knowing the pz code.

mbarton:
Fixes #185 

Normal users only see PDUs they belong to in the view preference switcher.

### DIFF
--- a/project/npda/general_functions/create_session.py
+++ b/project/npda/general_functions/create_session.py
@@ -6,13 +6,15 @@ from project.npda.general_functions import (
 from project.npda.models.organisation_employer import OrganisationEmployer
 
 
-def create_session_object_from_organisation_employer(
-    organisation_employer: OrganisationEmployer,
-) -> dict:
+def create_session_object_from_organisation_employers(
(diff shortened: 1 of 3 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 171 hours, 5 minutes and 15 seconds ago) Please check your changes!

Issue #189, opened 2024-07-15
Normal users can construct malicious request to see all data


mbarton:
As part of testing #185 I've noticed we don't perform permissions checking on setting `view_preference`. So even though a normal user can only see the PDUs they belong to in the switcher and can only see data for those PDUs, they can manually construct a request to set their `view_preference=2` and see all the data.

Easiest to reproduce in Firefox as it as a UI in the network inspector for editing and resending requests.

1) Create a user with a single organisation as their employer
2) Upload patient data against a different PDU
3) Log in as the first user
4) Click the Switcher button to trigger a view_preference change `POST` request
5) Right click on that request in the network inspector and click "Edit and resend"
6) Change the body of the request to say "view_preference=2"
7) Resend the request

**Expected**: the request doesn't work and you continue to see data just for your PDUs
**Actual**: You see all the data :(

https://github.com/user-attachments/assets/673411a1-01c7-4d27-9dbd-3f5589db4648

Issue #180, opened 2024-07-11
ods_code and pz_code not set in session if logging in from a given page


mbarton:
To reproduce:

1) Ensure you're logged out
2) Visit `/patients`
3) Log in

Once redirected the switcher should show your PDU but doesn't show anything.

![Screenshot from 2024-07-11 14-45-57](https://github.com/rcpch/national-paediatric-diabetes-audit/assets/395805/f0b8c2ba-3c8e-49ec-9de5-599433d8da17)

mbarton:
It seems like response_url is not login_redirect_url here: https://github.com/rcpch/national-paediatric-diabetes-audit/blob/b227bf35c7aefa732cd0a9c148107f54fa233ee3/project/npda/views/npda_users.py#L519

For example clearing cookies and trying to reload the `/patients` page:

```
!! done: response_url: /patients login_redirect_url: /account/two_factor/
```

I'm having to clear cookies due to the log out button dropdown not appearing (#181)

mbarton:
Fixed in #190

PR #190, opened 2024-07-16
Set session data in user_logged_in signal

mbarton:
The previous method of swapping the sessionid in client cookies didn't seem to work so before this PR the tests weren't actually logged in.

We worked around that with a check for either a 200 **or** a 302 code and inspecting the querysets directly but that won't work in future tests that send POST requests (eg tests for #189).

I've changed the session data to be added in the user_logged_in signal. That way we don't need the session swapping test code at all. I've tested tht the signal is not fired until two factor authentication is complete. It has the nice side benefit of also fixing https://github.com/rcpch/national-paediatric-diabetes-audit/issues/180

I've also changed the tests to check the actual QuerySet passed through to the template via context. This is closer in spirit to an end-to-end test and doesn't require any manual setup or calls to get_queryset.

### DIFF
--- a/project/npda/signals.py
+++ b/project/npda/signals.py
@@ -11,6 +11,9 @@ from django.dispatch import receiver
 
 # RCPCH
 from .models import VisitActivity, NPDAUser
+from .general_functions.create_session import (
+    create_session_object
+)
 
 # Logging setup
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 146 hours, 32 minutes and 15 seconds ago) Please check your changes!

PR #191, opened 2024-07-16
Check permissions when changing filter options from the view_preference switcher

### LINKED ISSUE #189: Normal users can construct malicious request to see all data

mbarton:
Fixes #189 where users could construct malicious requests allowing them to set any view_preference they wanted. This PR now applies permissions checking immediately for organisation, PDU and national view preference.

I've added some tests but only for the users view so far as we don't yet have patient seeding functions for testing. This PR however refactors out the checks to common code shared between the two views so applies to both.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 5 minutes and 36 seconds ago) Please check your changes!

Issue #192, opened 2024-07-16
Uploading the dummy sheet ends up with patients failing validation


mbarton:
![Screenshot 2024-07-16 15 42 48](https://github.com/user-attachments/assets/9e05b4dc-c247-4c2a-84bf-94a9f7e37016)
![Screenshot 2024-07-16 15 43 02](https://github.com/user-attachments/assets/1f5b5e0d-336a-4222-b53c-b9a96af6318b)

mbarton:
Um yeah so this happens because it doesn't matter what CSV you upload, we're going to process `dummy_csv_invalid.csv` anyway!

https://github.com/rcpch/national-paediatric-diabetes-audit/blob/1063ff66feecc69fe1c8f6134f34c219bd0fba8c/project/npda/general_functions/csv_upload.py#L77

eatyourpeas:
Haha sorry my bad. Was a shortcut for the demo and forgot to raise an issue
for it

mbarton:
https://github.com/rcpch/national-paediatric-diabetes-audit/pull/193 fixes it to use the provided file but there are still validation errors per row. I'll address in a follow up PR

mbarton:
Fixed in https://github.com/rcpch/national-paediatric-diabetes-audit/pull/196

PR #193, opened 2024-07-17
Fix CSV upload

mbarton:
Partial fix for #192

- Remove hard-coded invalid CSV and actually process the uploaded CSV file :)
- Seek the file back to zero to fix "no columns" error on the second call as you can't read the same file twice!


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 124 hours, 45 minutes and 26 seconds ago) Please check your changes!

PR #196, opened 2024-07-17
Fix validation errors uploading dummy sheet

mbarton:
Now when you upload `dummy_sheet.csv` you get a nice clean set of patients and visits with no errors.

- It had true/false for some columns instead of 1/2/99 as per the [dataset spec](https://www.rcpch.ac.uk/sites/default/files/2024-02/npda_dataset_2021_guidance_updated_feb_2024.pdf)
  - I think it's reasonable for us to handle true or false too but want to ensure that `dummy_sheet.csv` is our gold standard test case
- Don't create validation errors for missing fields
  - You would go to edit them in the UI but then they would magically fix themselves as it the errors were only produced during CSV upload


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 122 hours and 54 minutes ago) Please check your changes!

Issue #47, opened 2024-05-13
New patient form validation errors


dc2007git:
When going through the patient form, submitting an empty list produces (as expected) "This field is required" errors.

Line 56 in patient_form.py begins the start of validating each field specified as below:

`        if diagnosis_date is None:
            raise ValidationError({'diagnosis_date': ["'Date of Diabetes Diagnosis' cannot be empty"]})
        
        if date_of_birth is None:
            raise ValidationError({'date_of_birth': ["'Date of Birth' cannot be empty"]})

        if diagnosis_date is not None and date_of_birth is not None:
            if diagnosis_date < date_of_birth:
                raise ValidationError({'diagnosis_date': ["'Date of Diabetes Diagnosis' cannot be before 'Date of Birth'"]})
        
        if death_date is not None and date_of_birth is not None:
            if death_date < date_of_birth:
                raise ValidationError({'death_date': ["'Death Date' cannot be before 'Date of Birth'"]})
        
        if death_date is not None and diagnosis_date is not None:
            if death_date < diagnosis_date:
                raise ValidationError({'death_date': ["'Death Date' cannot be before 'Date of Diabetes Diagnosis'"]})`

The form then produces an error if date_of_birth or diagnosis_date is empty. But ValidationError is an Exception so will exit the function when the first conditional in the validation sequence is logged. This means that other all errors would not be listed. 

Just a small aside - we probably don't need the two first conditionals as django will produce a "This field is required" error if their conditions are met, so we'd double up on error messages!

eatyourpeas:
Absolutely right @dc2007git good spot. Django can accept multiple fields to the ValidationError class, so we could collect these up and raise at the end of all the conditionals? Would you be happy to look at and put in a PR?

Issue #31, opened 2024-05-05
Share validation functions between CSV upload and edit forms


eatyourpeas:
Allied to #24 
The API relating to to visits will need access to the validation functions used in the csv_upload

mbarton:
We should be able to re-use the `PatientForm` and `VisitForm`, manually binding data from the CSV row, call `.clean()` and check `errors` on the form.

We'll need to implement #47 to collect all the errors in one go rather than raising exceptions.

mbarton:
Both CSV upload and form submission now use the same validation code. When we introduce an API in the future we'll need to look at this again, although the shortest route to something usable is probably to submit the existing form behind the scenes in the API as we do with CSV upload

PR #200, opened 2024-07-18
Faster CSV uploads

### LINKED ISSUE #147: Validating each postcode on a CSV upload takes multiple seconds
For each row in the CSV upload we validate the postcode against https://findthatpostcode.uk. This takes ~0.8 seconds per postcode meaning the time to process the CSV is roughly the number of rows in seconds. This feels quite slow to the user.

We look up the index of multiple deprivation for each postcode via our [rcpch-census-platform](https://github.com/rcpch/rcpch-census-platform) API. This again validates the postcode but using postcodes.io instead.

postcodes.io allows bulk validation of up to 100 postcodes. We call it via our [rcpch-census-platform](https://github.com/rcpch/rcpch-census-platform) API though so we'd need to add a bulk endpoint to that too.

We switched to findmypostcode in E12 as postcodes.io was missing some postcodes. (https://github.com/rcpch/rcpch-audit-engine/issues/558, https://github.com/rcpch/rcpch-audit-engine/pull/576). At the moment if the postcodes.io lookup fails (via the IMD call) then you'll just end up with no IMD. I'm not sure of the consequences of switching everything back. I don't believe findthatpostcode offer a bulk endpoint.

### LINKED ISSUE #148: Getting the GP details takes multiple seconds during CSV upload
Similar to #147

Calling https://github.com/rcpch/rcpch-nhs-organisations to get the GP details for the ODS code takes ~0.5 seconds per row. This adds up across the entire CSV file to make the process feel slower.

Since that's our API we could add a bulk lookup endpoint to speed things up.

mbarton:
Fixes #148
Fixes #147

Make processing CSV uploads an order of magnitude faster by making all the external API requests in parallel across all rows.

I've done this by making the parent view async along with all the csv_upload functions and the `validate_postcode` and `imd_for_postcode` third party API lookups. The CSV upload code then creates a Python TaskGroup and submits a task to process each row.

The tasks share an [httpx](https://www.python-httpx.org/) [AsyncClient](https://www.python-httpx.org/async/) so they can co-operatively submit and wait for the responses. For the existing sync calling code on patient create and update I've added wrappers to create a client on demand for the duration of the requests.

I got the approach from this blog post: https://fly.io/django-beats/running-tasks-concurrently-in-django-asynchronous-views/. It's my first time with async Python though so it's very possible I've done something wrong or inefficiently. In particular this approach is naive in proportion to the number of rows in the file. Submitting a CSV with thousands of rows will launch thousands of requests simulataneously which is almost certainly not the behaviour we want. That said, this speeds up our current demo with `dummy_sheet.csv` so much I think it's probably worth merging as is and iterating on. Async code is fiddly by nature so we should not adopt this approach across the board, only where we deem it necessary.

Time to upload `dummy_sheet.csv`:

- *Before*: 11.91 seconds
- *After*: 3.01 seconds! 🚀

### DIFF
--- a/project/npda/general_functions/csv_upload.py
+++ b/project/npda/general_functions/csv_upload.py
@@ -3,12 +3,15 @@ from datetime import date
 import time
 import logging
 import os
+import asyncio
 from typing import Literal
 
 # django imports
 from django.apps import apps
 from django.conf import settings
 from django.utils import timezone
+
+from asgiref.sync import async_to_sync, sync_to_async
 
 # third part imports
 import pandas as pd
@@ -35,16 +38,16 @@ from ...constants import (
     DKA_ADDITIONAL_THERAPIES,
 )
 
-from .validate_postcode import validate_postcode
+from .validate_postcode import avalidate_postcode
 from .nhs_ods_requests import gp_details_for_ods_code
 from .quarter_for_date import retrieve_quarter_for_date
-from .index_multiple_deprivation import imd_for_postcode
+from .index_multiple_deprivation import aimd_for_postcode
 
 # Logging setup
 logger = logging.getLogger(__name__)
 
 
-def csv_upload(user, csv_file=None, organisation_ods_code=None, pdu_pz_code=None):
+async def csv_upload(user, csv_file=None, organisation_ods_code=None, pdu_pz_code=None):
     """
     Processes standardised NPDA csv file and persists results in NPDA tables
 
@@ -58,15 +61,15 @@ def csv_upload(user, csv_file=None, organisation_ods_code=None, pdu_pz_code=None
     AuditCohort = apps.get_model("npda", "AuditCohort")
 
     # set previous quarter to inactive
-    AuditCohort.objects.filter(
+    await AuditCohort.objects.filter(
         pz_code=pdu_pz_code,
         ods_code=organisation_ods_code,
         audit_year=date.today().year,
         quarter=retrieve_quarter_for_date(date_instance=date.today()),
-    ).update(submission_active=False)
+    ).aupdate(submission_active=False)
 
     # create new quarter
-    new_cohort = AuditCohort.objects.create(
+    new_cohort = await AuditCohort.objects.acreate(
         pz_code=pdu_pz_code,
         ods_code=organisation_ods_code,
         audit_year=date.today().year,
@@ -82,7 +85,7 @@ def csv_upload(user, csv_file=None, organisation_ods_code=None, pdu_pz_code=None
     except ValueError as error:
         return {"status": 500, "errors": f"Invalid file: {error}"}
 
-    def validate_row(row):
+    async def validate_row(row):
         """
         Validate each row
 
@@ -94,6 +97,8 @@ def csv_upload(user, csv_file=None, organisation_ods_code=None, pdu_pz_code=None
         """ 
         get all the data for this row
         """
+        print(f"!! validate_row: {row}")
+
         # Patient fields
         row_number = row["NHS Number"].replace(" ", "")
         date_of_birth = row["Date of Birth"]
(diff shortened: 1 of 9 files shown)


project/npda/general_functions/csv_upload.py
```diff
@@ -57,15 +72,15 @@ def csv_upload(user, csv_file=None, organisation_ods_code=None, pdu_pz_code=None
     AuditCohort = apps.get_model("npda", "AuditCohort")
 
     # set previous quarter to inactive
-    AuditCohort.objects.filter(
+    await AuditCohort.objects.filter(
```
mbarton, line 75: since Django provides async versions of the ORM functionality I've simply used that where possible to avoid as much sync_to_async wrapping (since that implies a bigger code refactor pulling out functions all over the place)

project/npda/general_functions/csv_upload.py
```diff
@@ -820,12 +836,19 @@ def save_row(row, timestamp, cohort_id):
 
         nhs_number = row["NHS Number"].replace(" ", "")
 
+        postcode = row["Postcode of usual address"]
+        
+        index_of_multiple_deprivation_quintile = None
```
mbarton, line 841: Moved from `Patient.save` as that hook can't be async

project/npda/general_functions/csv_upload.py
```diff
@@ -1073,17 +1091,20 @@ def save_row(row, timestamp, cohort_id):
...
+                        tg.create_task(save_row(row, timestamp, new_cohort.pk, client))
     except Exception as error:
         # There was an error saving one or more records - this is likely not a problem with the data passed in
+        logger.error(f"Error during csv upload {error}")
         return {"status": 500, "errors": error}
 
     return {"status": 200, "errors": None}
 
-
+@sync_to_async
```
mbarton, line 1107: Probably trivial to make the calls within this function async but no need too

project/npda/general_functions/index_multiple_deprivation.py
```diff
@@ -4,34 +4,50 @@
 
 # Standard imports
 import logging
-import requests
+import httpx
```
mbarton, line 7: The [fly.io blog post](https://fly.io/django-beats/running-tasks-concurrently-in-django-asynchronous-views/) suggested [httpx](https://www.python-httpx.org) and it looks great and is requests API compatible. I have no strong preference what we use though so long as it supports async and sadly httpx does bring [some transitive dependencies](https://www.python-httpx.org/#dependencies).

project/npda/views/home.py
```diff
@@ -20,26 +22,29 @@
 # Logging
 logger = logging.getLogger(__name__)
 
+
+@sync_to_async
 @login_and_otp_required()
```
mbarton, line 26: This decorator isn't async compatible so you can fudge it by wrapping then unwrapping. It does break our login logging though (see the TODO MRB in decorators.py above). Django does [allow login_required to be applied to async views](https://code.djangoproject.com/ticket/31949) so it's probably just a case of changing our decorator to be written in the same way

project/npda/views/patient.py
```diff
@@ -206,6 +206,10 @@ def form_valid(self, form: BaseForm) -> HttpResponse:
         )
         patient.site = site
         patient.is_valid = True
+
+        if patient.postcode:
```
mbarton, line 210: As above, needed to come out of `Patient.save` for async support.

project/npda/views/patient.py
```diff
@@ -240,6 +244,12 @@ def get_context_data(self, **kwargs):
     def form_valid(self, form: BaseForm) -> HttpResponse:
         patient = form.save(commit=False)
         patient.is_valid = True
+
+        if patient.postcode:
+            imd = imd_for_postcode(patient.postcode)
+            if imd:
```
mbarton, line 250: Sneaky bug fix, don't overwrite an IMD score if the API transiently returns an error

eatyourpeas (approved):
this works for me - a bunch quicker

mbarton:
To properly integrate this after #212 I'll need to refactor the CSV validation functions to the point it makes sense to do #31 and #47 first before rebuilding this PR on top of `live` again

PR #208, opened 2024-07-22
Add git metadata and prout config

mbarton:
https://github.com/guardian/prout

Will automatically track whether PRs make it out to production (obv just staging for now).

PR #211, opened 2024-07-22
Embed build info into our image

mbarton:
The previous attempt in #208 worked locally but not in the container because .git is ignored in our .dockerignore file. I've changed the whole process to write a cache file in CI which is burnt into the image.

Build info is required to enable Prout on this project: https://github.com/guardian/prout

When running locally there is no cache file but for simplicity it calls the same bash script to get the version info. In all cases the build info is then cached in a global variable to avoid extra i/o on every request.

Unfortunately creating the build info does need to be in a separate script that can be run from the host as the dockerignore means we will never have git information in the container. That means we can't add creating the build info to the Dockerfile.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 5 minutes and 3 seconds ago) Please check your changes!

Issue #21, opened 2024-05-05
403/404/500 custom pages


eatyourpeas:
Need custom templates and redirects for 403, 404 and 500 errors

PR #212, opened 2024-07-22
Fix duplicate patients with multiple visits in CSV upload

### LINKED ISSUE #206: Multiple visits for the same patient in CSV upload causes duplicate entries
Duplicate the final row in `dummy_sheet.csv` and change the visit date.

- *Expected*: you see one patient with two visits in the patients tab
- *Actual*: you see two patients with the same NHS number, each with one visit

mbarton:
Fixes #206 

Correctly handle patients with multiple visits in a CSV upload by grouping by NHS number and only adding the patient once.


project/npda/general_functions/csv_upload.py
```diff
@@ -821,22 +821,33 @@ def save_row(row, timestamp, cohort_id):
...
-                gp_practice_ods_code=row["GP Practice Code"],
-                is_valid=patient_is_valid,
-                errors=(patient_errors if patient_errors is not None else None),
-            )
+                "gp_practice_ods_code": row["GP Practice Code"],
+                "is_valid": patient_is_valid,
+                "errors": (patient_errors if patient_errors is not None else None)
+            }
+
+            # We can't use update_or_create as we need to ask AuditCohort if there is already a patient
```
mbarton: This feels like a constraint we should enforce at the database level. I'm scared that even with logging telling us if `existing_patient.count() > 0` we wouldn't notice until the update had updated some things we didn't want it to.

From a quick look (and probably missing a lot of context!) it feels like the relationship between `AuditCohort` and `Patient` should be held on the `Patient` and not the `AuditContext`. We could then use `update_or_create`.

eatyourpeas: One patient can be in many AuditCohorts, and a given AuditCohort can have many patients - hence the many to many relationship. One cohort will have 4 quarters so potentially will have the same patient 4 times. A previous requirement was that we store all csv uploads so currently a given quarter can have the same patient more than once, except on upload of an existing patient to a given cohort, the previous record has the `submission_active` flag set to false and the new one set to true, so that each quarter can only have one active submission per patient. I am not sure how necessary this is, but it was something Holly insisted on in the beginning so is currently implemented.
Before I close this PR, perhaps we need a conversation about data model in general - I am keen not to overcomplicate this and perhaps you and @anchit-chandran can stop me from doing this?

mbarton: Agreed we should chat. I've written up my thoughts in the forum (RCPCH internal): https://forum.rcpch.tech/t/submission-design/154. Broadly I wonder if we should split `AuditCohort` into two: `SubmissionPeriod` and `Submission`

mbarton:
I'm thinking it's probably easier to break out creating patients to a separate function - so we group the input CSV by NHS number, create the patients then create all the visits. That would still be async compatible with #200 and doesn't require any decisions on data modelling as uploading a CSV will always create a new submission anyway.

Marking this PR as DO NOT MERGE until I've made those changes

project/npda/general_functions/csv_upload.py
mbarton: Resolving as this PR no longer looks at AuditCohort - the conversation is ongoing though

project/npda/general_functions/csv_upload.py
```diff
@@ -782,311 +782,293 @@ def validate_date(
...
-                reason_leaving_service=(
-                    row["Reason for leaving service"]
-                    if not pd.isnull(row["Reason for leaving service"])
-                    else None
-                ),
-                paediatric_diabetes_unit_pz_code=pdu_pz_code,
-                organisation_ods_code=organisation_ods_code,
-            )
-        except Exception as error:
-            raise Exception(f"Could not save site: {error}")
```
mbarton, line 819: These try/catches just tended to hide the stack trace during development which was very annoying so I've removed them. We may want to add them back to avoid an angry blank screen, but tbh custom error pages are probably the better way to go (https://github.com/rcpch/national-paediatric-diabetes-audit/issues/21) so that we are sure we will get full stack trace logging in the backend logs

project/npda/general_functions/csv_upload.py
```diff
@@ -782,311 +782,293 @@ def validate_date(
...
-                "thyroid_function_date": (
-                    row["Observation Date: Thyroid Function"]
-                    if not pd.isnull(row["Observation Date: Thyroid Function"])
-                    else None
-                ),
-                "thyroid_treatment_status": (
+                )
+                else None
+            ),
+            "glucose_monitoring": row[
```
mbarton, line 875: This code is the same as before, just at one level of indentation lower. It was tempting to do some additional refactoring here but I'm leaving it for #31

eatyourpeas (approved):
this is working for me.

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 5 minutes and 30 seconds ago) Please check your changes!

PR #223, opened 2024-07-28
Add-remove-organisation-button-npdauser-form

eatyourpeas:
### Overview
Slightly more complicated PR than anticipated.
The refactor of models to create OrganisationEmployer, PaediatricDiabetesUnit models meant a restruction of the user update and creation workflow.
Since a user can have more than one employer, this workflow is handled with htmx in the npdauser form.
Now the user form lists all affiliated PDUs and allows the user to add/delete this list and set which is the primary organisation, based on permissions. There is some latency with this as there are API calls, so it would be worth putting a spinner on button press to signpost this.
In the user edit form superusers see a national list of organisations with their own removed, whereas normal users see only those organisations in their own PDU, with their own removed. In the create userform, the lists include their own organisations.

### Code changes
- `npdauser_form` - includes 2 partials - `employers.html` which has lists of affiliated organisations with htmx buttons to delete, add or assign primary PDU: `rcpch_organisations_select.html` - a select for the orgs
- `views/npda_users`: removes a lot of API look up which is abstracted out to the adapter in the createuser and updateuser classes.
-`admin.py`: incidental unrelated change. added PaediatricDiabetesUnit to admin
- `organisations_adapter.py` has function which accepts a user instance and a request. Uses these to return lists for the rcpch organisations selects in the update/create forms.
-`tailwind.css`: adds a new hidden class which is used to toggle the add / hide organisations select in the user form using hyperscript

### Documentation changes (done or required as a to follow


anchit-chandran:
Reviewed with @eatyourpeas. all happy

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @eatyourpeas and merged by @anchit-chandran 5 minutes and 6 seconds ago) Please check your changes!

Issue #219, opened 2024-07-27
Refactor tests as per new data model


anchit-chandran:
After changing the data model (#218 ), tests will be expected to break - will be a case of changing method names etc.

PR #227, opened 2024-07-31
Fixing-tests-from-new-datamodel

anchit-chandran:

### Overview
We have a new data model, thus tests need updating to reflect this, including the creation / refactoring of the various Factories.

### Code changes
- refactors to all tests / factories
- refactor `get_single_pdu_from_ods_code` to return a `PDUWithOrganisations` object with error values, in the case of an error, instead of just a normal dict. Will refactor all functions in this module fully in a separate PR

- nil

### Related Issues
- #219


anchit-chandran:
Discussed with @eatyourpeas just now. All happy

PR #241, opened 2024-08-13
active-selection-don't-delete

### LINKED ISSUE #201: Don't allow the currently active submission to be deleted
I can't think of a reason to do this other than by accident?

eatyourpeas:
### Overview
What should happen
The active submission in the submission table should have a disabled delete button next to it. If a delete post request is made, the submission should not delete and an error message should be returned to the template. If a submission is deleted at database level, this should through a 500 error if submission_active is true
What actually happens
The submission and all associated patient data is deleted

### Code changes
A guard statement is added to the Submission model in the delete method. This should never get called.
A check is added to the post method in `SubmissionListView` in the htmx callback to check if the submission instance is active and if so raise and error message.
Finally the submission_history.html template disables the delete button where `submission_active` is true

This should prevent users at all levels from deleting active submissions.

TODO: test(s) needs writing for this. I will add it to the list

### Related Issues
closes #201

### Mentions


eatyourpeas:
I have add some small UI bits - the invalid rows in the visit table did not match up with the header so that is fixed
There is a clarification box in the data quality report.
This still needs some work but good enough here to merge and I will raise a separate issue
image

eatyourpeas:
This needs merging before work starts on #243

project/npda/views/home.py
```diff
@@ -44,29 +42,38 @@ def home(request):
...
         file = request.FILES["csv_upload"]
         pz_code = request.session.get("pz_code")
-        
+
         summary = csv_summarise(csv_file=file)
 
         # You can't read the same file twice without resetting it
         file.seek(0)
-        
+        print(file.name, file.size)
```
mbarton: Remove this

mbarton (commented):
Looks good, just a rogue print statement

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 5 minutes and 22 seconds ago) Please check your changes!

PR #242, opened 2024-08-13
Periodically get a new Azure DB token

### LINKED ISSUE #207: Deployed NPDA loses connection to the database after a few days
We need to port across this fix from E12: https://github.com/rcpch/rcpch-audit-engine/pull/1004

It's not obvious how to to run this in Azure Container Apps. There are [scheduled jobs](https://learn.microsoft.com/en-us/azure/container-apps/jobs?tabs=azure-cli) which could write the token to the same Azure file share we store our config. But you configure them totally separately from the apps so the update job would have it's own managed identity and the main app would end up using a token from a different managed identity. They could perhaps share a user managed identity?

We could alternatively update the token on a cron within the docker container or a thread within the app.

mbarton:
Fixes #207.

Periodically get a new Azure DB token. Do it in a background thread as we don't have a sensible cron setup within the container. Azure Container Apps does have scheduled jobs but they run as totally separate managed identities so it would get confusing when monitoring things that the main app uses a token from an identity other than it's own. This solution works regardless of how we deploy the app.

We still get a token as the first step in the `start-prod` script.

The tokens live for a day but 15 minutes seems like a reasonable amount of time to avoid doing it too much. From what I can tell the token is cached somewhere within Azure Container Apps anyway. I guess *really* we should drive this period from the expiry time on the token itself but nothing else in the app parses it yet so this seems a reasonable compromise.

### DIFF
--- a/project/npda/apps.py
+++ b/project/npda/apps.py
@@ -1,5 +1,22 @@
+import threading
+
 from django.apps import AppConfig
 
+from .management.commands.write_azure_pg_password_file import write_azure_pg_password_file 
+
+def periodically_update_azure_pg_password_file():
+    update_interval = 60 * 15 # 15 minutes
+
+    def _periodically_update_azure_pg_password_file():
+        write_azure_pg_password_file()
+        
+        thread = threading.Timer(update_interval, _periodically_update_azure_pg_password_file)
+        thread.daemon = True
+        thread.start()
+    
+    thread = threading.Timer(update_interval, _periodically_update_azure_pg_password_file)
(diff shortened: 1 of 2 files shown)


mbarton:
Overdue on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 15 minutes and 3 seconds ago) What's gone wrong?

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 45 minutes and 34 seconds ago) Please check your changes!

Issue #236, opened 2024-08-10
`view_preference` does not default to checked


eatyourpeas:
Allied to #232 
What should happen
The`view_preference` toggle should show checked on page load to the user's `view_preference` value.  
What happens
The lists correctly filter to the `view_preference` but the toggle shows as unchecked (blue)

PR #248, opened 2024-08-18
Add PDU selector to navbar

### LINKED ISSUE #232: restyle view preference and move to navbar
The view context scopes all submissions, patients and visits to the PDU. If Currently there is a select for PDUs viewable to the user which is a partial that is reused in the listviews.

It would look and feel nicer if the selector was moved into the navbar and made smaller.

dc2007git:
So far, the navbar has been restyled with DaisyUI's navbar component.

Changes:

- The navbar's options have now been equally spaced
- The selected navbar option is persisted with rcpch_light_blue font, instead of borders which when hovered over would change the size of the navbar
- Remove empty link in navbar for unauthenticated users for 'Request Account'
- Add view preference selector to navbar

 
Navbar in development:
image

Work-in-progress navbar in this branch:
image

Still to be done:
- [x] NPDA vs National Paediatric Diabetes Audit in header
- [ ] as in issue #236 , view_preference doesn't default to checked when a new session has been signed into. For a navbar element that is always seen, this should definitely be fixed
- [ ] When on the 'submissions' page, the view_preference selector doesn't want to be changed. This makes sense as there was never a selector on the page before it was introduced in the navbar. Once this is fixed then the view_preference context will persist successfully between patients, users and submissions
- [ ] Implement consistent styling between RCPCH navbar and that of NPDA
- [ ] When on 'users' page and selecting 'All' in navbar PDU selector, after clicking Paediatric Diabetes Unit to select a new PDU, the dropdown doesn't load until the page is refreshed.
- [ ] Upon selecting a new PDU, the 'Patients under the care of {pdu}' banner does not update with the newly selected PDU, again until the page is refreshed. This bug seems to be present in live as well.


eatyourpeas (approved):
This all looks good Danny
I will merge in and try and incorporate with the work doing currently on refactoring out ods code etc

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @dc2007git and merged by @eatyourpeas 5 minutes and 22 seconds ago) Please check your changes!

PR #250, opened 2024-08-18
Ods-pz-refactor

eatyourpeas:
### Overview
A large refactor with not everything working - especially view preference and filtering

Maint things are to:
1. Move the PDU list to the PaediatricDiabetesUnit model: this sees parent and PDU names and lead organisation ods code and name. There is one for each in the UK and Jersey. RCPCH is included. It is seeded in the s/start-dev and has a function in manage.py to seed.
2. Implements a new endpoint from RCPCH NHS Organisations API to pull a list of PDUS (including RCPCH) with their parent and lead organisations and trust/health boards. There is as well as a function to compare with the API list and update accordingly - in future this will be run on a schedule.
3. Refactors the session object now to store only the pz_code. This is used draw the lists for the selects in the forms and the `view_progress`. These are now pz code and lead organisation and parent names.
4. ods code is nolonger used anywhere in the app for filtering and the sibling organisations concept has been removed. The API endpoint is still live
5. forms (esp NPDAUser) have been much simplified and duplicated logic removed. The forms handle no saving or select instantiation logic - this is all handled or passed in from the view.

Implications
The view preference does update the session but now does not remain set at the correct PDU. 
The csv_upload validates appropriately but now adds an item that is valid in the invalid list


project/npda/general_functions/organisations_adapter.py
```diff
@@ -82,3 +88,105 @@ def organisations_to_populate_select_field(request, user_instance=None):
...
+            or request.user.is_rcpch_audit_team_member
+            or request.user.is_rcpch_staff
+        ):
+            # return all paediatric diabetes units excluding those were the user is employed
+            return (
+                PaediatricDiabetesUnit.objects.all()
+                .exclude(npdauser=request.user)
+                .order_by("organisation_name")
+                .annotate(
+                    paediatric_diabetes_unit_name=Concat(
```
mbarton: I'm not normally too hung up on DRY code but I think perhaps a function to build each of the annotate arguments would be useful to highlight what the differences are between each one?

It's hard to eyeball each one individually and understand what it's doing differently to the others

project/npda/general_functions/rcpch_nhs_organisations.py
```diff
@@ -131,116 +135,102 @@ def get_all_nhs_organisations_affiliated_with_paediatric_diabetes_unit() -> (
...
+        List[Tuple[str, str]]: A list of tuples containing the ODS code and name of NHS organisations.
+    """
+    url = f"{settings.RCPCH_NHS_ORGANISATIONS_API_URL}/paediatric_diabetes_units/trust"
+    ERROR_RESPONSE = [("999", "An error occurred while fetching NHS organisations.")]
+
+    try:
+        response = requests.get(url=url, timeout=10)  # times out after 10 seconds
+        response.raise_for_status()
+
+        if seed:
```
mbarton: I think this function is only called with `seed=True`. Should we remove the argument entirely and the code below this guard?

project/npda/general_functions/rcpch_nhs_organisations.py
```diff
@@ -131,116 +135,102 @@ def get_all_nhs_organisations_affiliated_with_paediatric_diabetes_unit() -> (
...
+                f"Created PaediatricDiabetesUnit: {new_pdu.pz_code} ({new_pdu.organisation_name})"
+            )
+        else:
+            if pdu_obj is not None:
+                if model_to_dict(pdu_obj) != model_to_dict(new_pdu):
+                    logger.info(
+                        f"{new_pdu.pz_code} ({new_pdu.organisation_name}) was updated."
+                    )
+                else:
+                    logger.info(
```
mbarton: I think this will be quite noisy in practice so probably one to remove before merging

project/npda/models/npda_user.py
```diff
@@ -104,37 +96,29 @@ def create_superuser(self, email, password, **extra_fields):
...
                 logged_in_user = self.create_user(
                     email.lower(), password, **extra_fields
                 )
-                paediatric_diabetes_unit, created = (
-                    PaediatricDiabetesUnit.objects.get_or_create(
-                        ods_code="RJZ01",
-                        pz_code="PZ215",
-                    )
+                paediatric_diabetes_unit = PaediatricDiabetesUnit.objects.get(
+                    pz_code="PZ215",  # Superusers that are not RCPCH staff are affiliated with King's College Hospital
```
mbarton, line 115: Would we have any superusers not attached to a PDU or the RCPCH?

project/npda/models/paediatric_diabetes_unit.py
```diff
@@ -10,15 +10,32 @@ class PaediatricDiabetesUnit(Model):
         max_length=10,
         help_text="Enter the paediatric diabetes unit PZ code",
     )
-    ods_code = CharField(
+    organisation_ods_code = CharField(
         max_length=10,
         help_text="Enter the organisation ODS code",
     )
+    organisation_name = CharField(
```
mbarton: Could we call these `lead_organisation` for clarity?

project/npda/general_functions/organisations_adapter.py
eatyourpeas: fair point. sorry sent this in possibly before it was finished. Not every select needs the same list of PDUs - that is what this trying to do. I will fix and let you know when ready to review.

project/npda/general_functions/rcpch_nhs_organisations.py
eatyourpeas: this is unlikely to be run more than weekly. it is polling the API and comparing the local list with what is there and updating as necessary. For sure can remove some of the logging. At the moment for some reason the logging is not working anyway.

project/npda/models/npda_user.py
eatyourpeas: no. I don't think so. I the group 4 are rcpch employees only. clinicians can have rcpch permissions but would have the is_rcpch_audit_team_member as true, unless you want us to remove that.

eatyourpeas (commented):
I have addressed all the issues above and now the drop down is working across the app. The only item you raised in the createsuperuser function I have left unchanged but happy to amend.

I now have 5 failing tests though, which are all centre on pytest login with the corresponding request not containing a user object. I think this is the way we set up a user in the factory and log them in but have spent some time on it and can't for the moment resolve.

I will chat with anchit to see if we can fix. just wondering whether we might merge this anyway now that we actually have some test users. Whatever you feel in any case

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 5 minutes and 23 seconds ago) Please check your changes!

Issue #243, opened 2024-08-14
Data quality report


eatyourpeas:
The data quality report is currently only visible on upload of a file. Once the user navigates to the submissions page, that information is lost.

The data quality report should therefore have it's own page or as a separate template visible below the submission list. 
in addition, the spreadsheet download button should pull down the uploaded data with an extra sheet including the report

- [x] save the csv data as a data field in the model on upload
- [x] move data quality report to render below the submission table for the active submission - should be a table with rows corresponding to csv rows, styled to look like excel or similar so user reassured it is the same date they uploaded.
- [x] download the submitted data via the download button
- [ ] include the DQR as a separate sheet in the downloadable csv.

mbarton:
Last task covered by https://github.com/rcpch/national-paediatric-diabetes-audit/issues/271

PR #257, opened 2024-08-21
data-quality-report-csv-download

eatyourpeas:
### Overview
1. Implements the Data Quality Report within submissions list
2. Fixes the lack of filtering of submissions against view preference
3. Renames view preference to PDU View and National View
4. Stores csv as a filefield in the Submission table
5. Updates settings.py, dockerfile and docker-compose to allow a media folder in the volumes to store csvs from the database
6. Implements CSV download button from submissions listview

### Code changes
- `Dockerfile` and `docker-compose`: add extra volume and media folder
- `urls.py` and `settings.py`: adds urls for csv storage in and download from media folder and sets up backend for same
- `csv_download`: new file to implement csv_download
- `csv_summarize`: correct spelling and pull into own file
- `submission.py`: add `csv_file` FileField in model and create migration for same
- `views/submissions.py`: tidy `SubmissionsListView` and fix queries based on `view_preference`, include data quality report in context
- templates: create new `data_quality_report` partial, include in `submission_history`. 
- add media folder to gitignore

This needs doing

### Related Issues
#243 - to complete this issue need still add data quality report as extra sheet prior to csv download. Otherwise all items complete.
Tests needed for this also


mbarton (changes requested):
Code looks fine and this will work locally but we haven't set up the Django `MEDIA_ROOT` to be persistent across our containers. So submissions will be lost after every deploy.

The quickest way to solve this would be to mount another Azure file share into the container (we already have one for the env file) and save into that.

I wonder though if using a [BinaryField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#binaryfield) is even simpler and keeps the implementation entirely within Django and the DB.

It's not ideal to store regularly accessed or updated files in the database but for something like our CSV submissions I don't see an issue with it - they're rarely written and rarely accessed.

Maybe let's discuss it IRL though - I'm not wedded to any option so long as it works within our CI pipeline

mbarton:
Sprint Meeting - 22/08/24: we'll use an Azure file share because it's the recommended way in Django. If we overwrite a file by accident we can always go back to the clinicians who uploaded it

mbarton (approved):

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @eatyourpeas and merged by @mbarton 5 minutes and 23 seconds ago) Please check your changes!

PR #258, opened 2024-08-21
Case insensitive email

### LINKED ISSUE #252: Email is case sensitive when creating new users
I created Cillian's email with capital letters (Cillian.Brophy@...). He was unable to log in, even if he used the exact case that I used. 
He was only able to log in once I changed his email to be all lower case. 

I think, in E12, emails were converted to all lowercase when a user was created.

mbarton:
Fixes #252 so you can log in with a case insensitive email. This is required since Microsoft products tend to change case depending on where they display the email address (in Teams, in the Outlook email compose fields etc).

In [Epilepsy 12](https://github.com/rcpch/rcpch-audit-engine) we implemented this by downcasing the email field in cleaners across each user form.

Django also used to have a model field specifically for this: `CIEmailField`. However they removed it in version 5. The recommended solution is to define a database collation to specifically configure how you want the case comparison to work, as detailed in this blog post: https://adamj.eu/tech/2023/02/23/migrate-django-postgresql-ci-fields-case-insensitive-collation/.

However as [this Django mailing list thread ](https://groups.google.com/g/django-developers/c/nDMnO98nexY) describes that method has unexpected interactions with other Django features.

So I've decided to use [django-citext](https://github.com/voiio/django-citext) which is a copy paste of the pre v5 code. We are unlikely to hit the special cases that require a full collation and by defining the case insensitive behaviour at the field level we won't cause bugs by forgetting a lowercase call later down the line in new features.


project/npda/migrations/0006_alter_npdauser_email.py
```diff
@@ -0,0 +1,30 @@
...
+
+    dependencies = [
+        (
+            "npda",
+            "0005_rename_organisation_name_paediatricdiabetesunit_lead_organisation_name_and_more",
+        ),
+    ]
+
+    operations = [
+        CITextExtension(),
```
mbarton: This will probably fail during CI as only the main admin user can create extensions in Azure Flexible Postgres. So I'll run `CREATE EXTENSION citext` manually on the database before deploying.

eatyourpeas:
Is there a reason we can't replicate what we do in E12 or is it because there is the possible situation where a user has capitalised names in their email address that the email provide might throughout the same email as lower case?

We have this already in E12 and NPDA: 
```
constraints = [
            UniqueConstraint(
                Lower("email"),
                name="user_email_ci_uniqueness",
            ),
        ]
```
I think this takes whatever email is entered by the user and downcases it for email comparison but stores it otherwise the way it was entered. I have misunderstood this bit.

mbarton:
> is it because there is the possible situation where a user has capitalised names in their email address that the email provide might throughout the same email as lower case?

Yeah in Microsoft products some places will return the email capitalized even if the user in practice uses it lowercase. By the RFC the part of an email address before the @ is case sensitive but in practice it's not for most providers, to avoid confusion and scam squatting on an existing address.

What happened to Amani and Cillian is she added his email capitalised and he couldn't log in until he used the exact same casing.

I think that constraint you posted will ensure that we can't have two entries in the database that match case insensitively but we would still need to remember to downcase the email address before passing it to Django to do authentication. The solution in this PR means that it doesn't matter what casing you use, you always get back the same user.

mbarton:
@eatyourpeas are you happy for me to merge this or shall we talk through the options?

eatyourpeas:
I think fine to merge

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 4 minutes and 20 seconds ago) Please check your changes!

PR #261, opened 2024-08-25
pagination-filter-sort

### LINKED ISSUE #202: Search and filter patients
We need to define what fields you can search on. I think a reasonable starter:

- NHS Number
  - Probably don't need partial search but do need to normalise the numbers (eg you can search with spaces)
- Date of Birth
- Postcode
  - As above, can probably start without partial search but normalising the spaces

Filters:

- Diabetes type

eatyourpeas:
### Overview
Implements pagination, filter by nhs number and npda unique id, sort by nhs number and npda unique id

### Code changes
- `patient_table.html` partial - added params to the get requests for the sorting and pagination buttons in the table footer
-`patients.html` - adds input field that sends input field value as param in hx-get request on keyup
- `patient.py` - add in the pagination functions to the patientlistview. alter the filters to accept the params from the GET requests (input box text or sort requests)

### Related Issues
closes #202


project/npda/templates/partials/patient_table.html
```diff
@@ -4,8 +4,92 @@ <h1 class="text-md font-montserrat font-semibold text-rcpch_dark_blue">Patients
 <table class="table table-md w-full text-sm text-left rtl:text-right text-gray-500 text-gray-400 mb-5 font-montserrat">
     <thead class="text-xs text-gray-700 uppercase bg-gray-50 bg-rcpch_dark_blue text-white">
             <tr>
-                <th scope="col" class="px-2 py-3 text-center">NPDA ID</th>
-                <th scope="col" class="px-2 py-3 text-center">NHS Number</th>
+                <th scope="col" class="px-2 py-3 text-center cursor-pointer">
+                        <span class="flex flex-row">
+                            <?xml version="1.0" ?>
```
mbarton: I wonder if we make a partial for each SVG icon just for readability?

project/npda/views/patient.py
```diff
@@ -40,43 +40,50 @@ class PatientListView(
     permission_denied_message = "You do not have the appropriate permissions to access this page/feature. Contact your Coordinator for assistance."
     model = Patient
     template_name = "patients.html"
+    paginate_by = 5
```
mbarton: This feels a bit low?

mbarton (changes requested):
Code looks good but I don't think the UX is right

Screenshot from 2024-08-28 12-29-24

Both the asc and desc icons have up and down icons so it's not obvious which one to click and you don't know which one is active.

We should highlight the active sorting order icon and have them only on one side of the column name

project/npda/templates/partials/patient_table.html
eatyourpeas: this is a great suggestion - I have done this.

eatyourpeas:
I agree thanks for this - how do we feel about this, or is it still too opaque. alternatively could just go for an up arrow and a down arrow
image

project/npda/views/patient.py
eatyourpeas: yes - sorry left over after testing. I will change to 50

mbarton:
> I agree thanks for this - how do we feel about this, or is it still too opaque. alternatively could just go for an up arrow and a down arrow

yes clicking the down arrow to sort descending and up for ascending would be how I’d expect it to work at a glance. No preference on whether it’s one icon or two

eatyourpeas:
finally settled on this
image

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 5 minutes and 21 seconds ago) Please check your changes!

PR #266, opened 2024-09-01
fake-patients

eatyourpeas:
### Overview
This PR is not actually ready for merging but thought I would use the mechanism to trigger code review before going any further.

The idea is to create a Child class with private methods that ultimately produce a Child instance that is identical largely to the Patient object but without a unique identifier (apart from `nhs_number`.)

To initialize the class the user passes in parameters they want to prescribe to create types of child (for example age category, sex and so on). The child instance will have randomized valid fields and this can be used to create a valid Patient instance, or individual fields can be changed to become invalid once created and these can then be passed in to the validation functions (when written) to create the errors and a final Patient object.

I hope that makes sense.

The first step to try it out would be to import the `data_generator` file in the shell and call:
`generate_valid_children(number_of_children=5)`

Equally you could just:
```console
>>> from project.npda.general_functions import *
>>> age_range = AgeRange.AGE_0_4
>>> child = Child(age_range=age_range)
>>> child.date_of_birth
datetime.date(2020, 11, 20)
```

If you are happy with the concept and general idea I will flesh out and create another one for visits and tie them all to transfer and submissions in a similar way.


mbarton (commented):
Looks fine to me, I'm not sure we need a class though rather than just having a function that returns a tuple of Patient, Transfer etc

anchit-chandran:
It looks fine to me but the factories do a lot of this - and the package has a lot more support for all the behaviours we'd need for doing this kind of thing. Is there a particular reason / benefit to roll our own here? Like one small thing is the  generated `nhs_number` is completely random, so could (low likelihood) clash with another one; but this is handled in the `PatientFactory` as so:

```python
class PatientFactory(factory.django.DjangoModelFactory):
    """Dependency factory for creating a minimum viable Patient.

    Fills in default values if not specified at object creation.
    """

    class Meta:
        model = Patient
        skip_postgeneration_save = True

    @factory.lazy_attribute
    def nhs_number(self):
        """Returns a unique NHS number which has not been used in the db yet."""

        # First generate 5 numbers and check if they exist in the db
        candidate_nums = nhs_number.generate(
            quantity=5, for_region=nhs_number.REGION_ENGLAND
        )

        not_found_unique_nhs_number = True
        while not_found_unique_nhs_number:
            for nhs_num in candidate_nums:
                if not Patient.objects.filter(nhs_number=nhs_num).exists():
                    not_found_unique_nhs_number = False
                    return nhs_num

            # If all 5 numbers are already in the db, generate 5 more
            candidate_nums = nhs_number.generate(
                quantity=5, for_region=nhs_number.REGION_ENGLAND
            )
```

eatyourpeas:
AH sorry @anchit-chandran I think I have misunderstood. I thought you/we needed a way of generating dummy records. This  branch does this and now I think works - the visit functions in particular generate sensible blood pressures/heights/weights etc against age ranges for example, and select measures appropriate for the diabetes type - eg will not score a child as being on an insulin pump if they have type II DM.

If you have factories that do this already I will leave this then - sorry.

mbarton:
I don't think the factories do everything in this branch. I'm implementing the "date of birth < 19" validator but the current date of birth in the factory makes a patient > 19.

I think @anchit-chandran is saying we should implement this branch on top of the existing factories rather than create new plain functions

eatyourpeas:
I think my code goes up to 25y - it generates a random date of birth based on a new AgeRange enum so that we can in generate children between preschool, primary school, early and late secondary and young adult. The audit technically stops on the 19th birthday but there is lots of interest in the young adults up to 25y. Probs I can just remove the upper age group if we are happy to keep this code. I don't want to introduce confusion or duplicate work though. Perhaps we need to chat before doing anything more

eatyourpeas:
just refreshing this conversation. is there any appetite anchit for me to try and include some of the visit stuff i have done into your factories? I would like us to be able to have a way of generating csvs to test @mbarton work on async with more than 10 kids in them. Also, if we denormalise nhs number so that we can have more than one record with the same nhs number, would be good to makes sure we can test carefully the situations in which we allow this

anchit-chandran:
Ready to merge. No shell script for running a realistic bunch of submission seeding yet.  

Adds:

# `class FakePatientCreator`

Required extending `Factories`.

Full documentation for use in mkdocs, with snippet below.

We initialise the FakePatientCreator with a specific audit period as
this is used throughout the creation of fake patients and visits
including:

- setting random date of birth so max age by audit_start_date is valid
    for the given AgeRange. Also, diagnosis_date is between
    date_of_birth and audit_start_date.

- setting random diagnosis date so it is before audit_start_date

- Visit dates are spread evenly throughout each quarter of the audit
    period. For each Visit, the date is randomly set within its
    quarter's date range.

```python
from project.npda.general_functions.data_generator_extended import (
    FakePatientCreator,
    HbA1cTargetRange,
    VisitType,
)

# Set necessary attributes to calibrate all dates
DATE_IN_AUDIT = date(2024, 4, 1)
audit_start_date, audit_end_date = get_audit_period_for_date(DATE_IN_AUDIT)

fake_patient_creator = FakePatientCreator(
    audit_start_date=audit_start_date,
    audit_end_date=audit_end_date,
)
```

The `FakePatientCreator` provides utility methods to generate and save test data for patients and their corresponding visits. This guide describes the usage of the following three key methods:

1. `build_fake_patients`: Builds fake patient objects.
2. `build_fake_visits`: Builds fake visit objects for patients.
3. `create_and_save_fake_patients`: Combines patient and visit creation and saves them to the database. Can pass in `patient_kwargs` and `visit_kwargs`.

# Tests for above

- Some basic tests made for generator

# `seed_submission` manage.py helper cmd

CLI wrapper around `FakePatientCreator` that additionally houses the created rows under a `Submission`.

Usage in docstrings `seed_submission.py`. Snippet:

```
CLI for seeding Submissions for a given User (and that user's PDU). Will
additionally create Patients and Visits.

Example use:
    python manage.py seed_submission \
        --pts=50 \
        --visits="CDCD DHPC ACDC CDCD" \
        --hb_target=T
        --user_pk=1 \
        --submission_date="2024-10-18" \

    This will generate a submission for User.pk=1 that includes 50 patients,
    each of whom will have 16 Visits evenly spread throughout the audit
    year's quarters, with types Clinic, Dietician, ..., Clinic, Dietician,
    all with HbA1c target range as TARGET.
```

mbarton:
Overdue on [STAGING](https://npda-staging.rcpch.tech) (created by @eatyourpeas and merged by @anchit-chandran 15 minutes and 4 seconds ago) What's gone wrong?

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @eatyourpeas and merged by @anchit-chandran 1 hour, 5 minutes and 38 seconds ago) Please check your changes!

PR #272, opened 2024-09-08
Move-media-store-submission-dates

eatyourpeas:
### Overview
Last sprint with the NPDA team the requirement was to retain all records of a submission upload, but only retain the csv file of the active submission.
It was also said that invalid records in the active submission should be retained and contribute to any scores. Messaging in the UI needs to reflect this
This PR:
- adds a versioned name to the uploaded csv file using pz code and timestamp
- stores the csv file in a media file outside the project folder but within the app folder
- allows csv_file to be empty so that inactive submissions have no csv_files attached
- in the upload process, if a previous submission becomes inactive, the csv_file is deleted and set to None in the database
- changes to the template have been made to reflect this

### Code changes
allows csv_file field to be null - new migration
deletes any csv files if submission_active false
retains record of upload in table
adds media folder to .gitignore
alters template to prevent download of old submissions
ere.

### unresolved issues
Note that if a submission is uploaded a second time, while the csv is removed, the patient and visit records remain in the tables. This means if information relating to a patient who was previously uploaded is uploaded again, a patient duplicate error is stored in the error field.
To overcome this we have two options:
1. to delete all patient/visit/transfer records in relation to the inactive submission (my preference)
2. leave the data as is but in the validations filter against submission as well as patient

Perhaps we need a conversation about this before creating an issue for the fix.


project/settings.py
```diff
@@ -123,6 +123,7 @@
 
 ROOT_URLCONF = "project.urls"
 
+# This directory is used to store the .csv files that are uploaded by the user
```
mbarton: dupe?

project/npda/general_functions/csv_upload.py
```diff
@@ -45,20 +45,55 @@ def csv_upload(user, csv_file=None, pdu_pz_code=None):
...
+            original_submission.csv_file = None
+            original_submission.save()
+
+    except Exception as e:
+        logger.error(f"Error creating new submission: {e}")
+        # the new submission was not created so we need to delete the csv file and
+        # reset the old submission to active and add the csv file back if a there was a previous submission
+        if original_submission:
+            original_submission.get().submission_active = True
+            original_submission.csv_file = new_submission.csv_file
```
mbarton: Wouldn't this mismatch the submissions?

We could also end up here if the original submission CSV file failed to delete - perhaps we need two separate try catches - one for creating the new submission and the other for errors during rollback (although the latter I guess could just be unhandled)

project/npda/general_functions/csv_upload.py
```diff
@@ -45,20 +45,55 @@ def csv_upload(user, csv_file=None, pdu_pz_code=None):
...
+            original_submission.save()
+        raise ValidationError(
+            {
+                "csv_upload": "Error creating new submission. The old submission has been restored."
+            }
+        )
 
     def csv_value_to_model_value(model_field, value):
```
eatyourpeas, line 112: my question here is - should I at this point delete all patients and visits and transfers associated with this submission, or leave them in place? If I do do that, I could either do that here, or in the save method in the model

project/npda/general_functions/csv_upload.py
eatyourpeas: You are right not sure where my brain took me there. The idea is that I don't want the clinicians ever to lose previously submitted data. I figured, if in the process of uploading a new csv there was an error/corrupted file etc, we needed to restore the existing record. By reordering and simplifying a little I have the same outcome in fewer lines.

eatyourpeas (commented):
> You are right not sure where my brain took me there. The idea is that I don't want the clinicians ever to lose previously submitted data. I figured, if in the process of uploading a new csv there was an error/corrupted file etc, we needed to restore the existing record. By reordering and simplifying a little I have the same outcome in fewer lines. 

update - conversation today @mbarton and @eatyourpeas. Decision to delete patients associated with an active submission

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 4 minutes and 27 seconds ago) Please check your changes!

PR #277, opened 2024-09-18
Additional patient validators and tests

### LINKED ISSUE #164: Can only look up GP in new patient form if postcode has a space
The NHS spine API is fussy and requires the postcode to have a space in it otherwise the lookup fails and you get a full screen error

### LINKED ISSUE #220: Add csv upload tests

mbarton:
- Port across the existing tests against the patient model to the patient form where the validators run currently
- Add tests for CSV upload (fixes #220)
- Re-arrange the code that modifies patient fields to follow what (I think) is Django best practice:
  - Run them in `to_python` on a subclass of field if they normalise the data (eg remove spaces from postcode)
  - Run them in `clean_XXXX` if they only validate a single field
  - Run them in `clean` if they need to validate a field using the values from others (eg diagnosis date not before dob)
  - Calculate the IMD overriding save on the form rather than on the model
- Only catch `RequestException` when wrapping remote network validators
  - That way we don't swallow logic errors but do swallow transient network errors
- Mock out all network calls in the unit test
  - I did this originally to mock in return values but it's good practice anyway and speeds up the tests too

This PR adds a couple of behaviour changes:

- Patients can't be older than 25
- Normalise the GP postcode before checking it (fixes #164)
- Store normalised postcode and GP postcode
- Validate the GP ODS code against the NHS spine
  - Before we just accepted it


project/npda/tests/form_tests/test_patient_form.py
```diff
@@ -0,0 +1,298 @@
...
+)
+
+# Logging
+logger = logging.getLogger(__name__)
+
+
+# We don't want to call remote services in unit tests
+@pytest.fixture(autouse=True)
+def mock_remote_calls():
+    with patch("project.npda.forms.patient_form.validate_postcode", Mock(return_value=True)):
```
mbarton: I tried reducing the nesting here with an [ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) but it didn't work. There's probably a way to do it but it's not that bad and wrapped up in a fixture anyway

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 5 minutes and 22 seconds ago) Please check your changes!

PR #286, opened 2024-09-24
Custom error pages

### LINKED ISSUE #21: 403/404/500 custom pages

reecehill:
### Overview
TBC.

### Code changes
TBC. 
nil.

### Related Issues
#21


eatyourpeas (commented):
@reecehill I have fixed the broken tests here - please check over to be happy. I put all the error tests into a single class and a single file: hope that is ok.

The permissions tests were breaking because the custom 403 did not explicitly set the error code to 403, so this was therefore returning 200. Hope that makes sense. I have only tested the 403 in the UI, not the 500 or 404 etc.

Also, @mbarton comment here - not sure about the wording in the templates. Phrases like `This may be due to a client error or a malformed request. ` is quite jargon-filled. 

Perhaps could simplify the wording for each? I agree though needs to be friendly and where possible offer users option to reach out to NPDA team if they are having difficulties. I wonder if we could get the team to do us a custom background for this, as they did in E12

I am puzzled though because the tests are now all passing for me locally but are still apparently failing on github.

eatyourpeas (commented):
Tests failing on github:
`Process completed with exit code 137.`
This means insufficient memory, so not sure why that would be.

mbarton:
@eatyourpeas @reecehill I think the out of memory issue happened because Django wouldn't start up. When I ran it locally I got:

```
django-1   | ERRORS:
django-1   | ?: (security.E101) The CSRF failure view 'project.npda.views.csrf_fail' does not take the correct number of arguments.
```

Adding `reason` to the arguments list seemed to fix it and after that it booted up fine:

```python
def csrf_fail(request, reason=""):
```

https://github.com/rcpch/national-paediatric-diabetes-audit/pull/286/files#diff-19a23cf44fc21774b2eb552016a6b0dec1367f4a8720efe8436c47d529d7158fR34

I guess since we launch docker compose in the background (`-d`) it was just spinning booting up until it ran out of memory?

reecehill:
Brilliant!!

Bizarre though, I didn't get this error locally at all.
Glad its sorted.

mbarton:
> Bizarre though, I didn't get this error locally at all.

Yeah weird - at the moment Django is an unpinned dependency so it's possible we ended up on different Django versions which caused the issue

mbarton:
> Also, @mbarton comment here - not sure about the wording in the templates. Phrases like This may be due to a client
> error or a malformed request. is quite jargon-filled.
>
> Perhaps could simplify the wording for each?

I've tweaked the wording a bit but we don't need to worry too much, people are fairly used to software spitting out strange messages when it stops working

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @reecehill and merged by @mbarton 6 minutes and 29 seconds ago) Please check your changes!

PR #298, opened 2024-10-05
Visit-form-tidy

eatyourpeas:
This fixes a bunch of UI issues in the visit form and the patient list

The patient list now highlights errors in the list relating to the patient in the affected field, and only signposts the visits with errors. The redness in the patient field which was overbearing is now constrained to the affected field.

In the visit form, the colours have been replaced as borders to the categories with small labels in each corner. There is a more subtle red outline for those categories with errors. The error labelling has the html escaped now and rendered on separate lines

There were issues with daisyui not importing the colours so there are changes to the config and dockerfile to make sure these are included

Maybe not quite there yet, but a step closer


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 5 minutes and 23 seconds ago) Please check your changes!

PR #303, opened 2024-10-06
Kpis-in-the-uis

eatyourpeas:
### Overview
First implementation of KPIs in the UI
It is not quite right though because the wording relates to the whole PDU, not the individual child
It should be possible for users to see if the child has passed or failed a given measure, so we will need to filter out those KPIs that are really aggregations.

This is a good start though - thank you Anchit

### DIFF
--- a/project/npda/kpi_class/__init__.py
+++ b/project/npda/kpi_class/__init__.py
@@ -1,0 +1,1 @@
+from .kpis import *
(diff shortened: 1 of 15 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 26 seconds ago) Please check your changes!

PR #306, opened 2024-10-06
Dashboard

eatyourpeas:
### Overview
This sets up for next step to aggregate across regions
The `CalculateKPIS` class now accepts a list of pz_codes rather than just one, and this will allow us to pass in a list of PDUs within a region and retrieve a summary of the kpis across them all.

It also creates a new page called `dashboard` which has a table of the KPIs for the selected PDU, and in future will have further levels of abstraction. This reloads on changes in the `view_progress`

### DIFF
--- a/project/npda/kpi_class/kpis.py
+++ b/project/npda/kpi_class/kpis.py
@@ -60,7 +60,7 @@ class CalculateKPIS:
         """Calculates KPIs for given pz_code
 
         Params:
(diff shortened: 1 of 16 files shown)


eatyourpeas:
image

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 5 minutes and 24 seconds ago) Please check your changes!

PR #324, opened 2024-10-17
Patients only unique by NHS number within submissions

### LINKED ISSUE #276: denormalize user
Currently if a user uploads a CSV, the rows in it are validated against the relevant model, errors collected of the affected fields and a record stored for each in relevant model. This works particularly for the visit model (which has no unique identifier) but not so much for the Patient model which complains if there are two instances of NHS number which currently has `unique=True` in the model.

If a user uploads the same patient in the same submission, the new csv will be saved, the old csv, patient and visit records are deleted and the new records saved. A record of the upload remains in the submission table, but it is inactive and all related csvs and patient data have been removed.

However, if someone else, in a different PDU creates a new submission which is active but contains the same patient, this will generate an error at database level (since NHS number is unique) and the submission will be saved but none of the patients.

This highlights a fundamental issue possibly about the data model and versioning of our records. One way round this would be to allow multiple records with the same NHS number in the Patient table, but not allow more than one record per Patient in the Transfer table in a given active submission. Possibly we need to chat about this.

eatyourpeas:
Fixes #276

Detailed conversation in the issue not repeated here.

This PR removes the unique attribute from nhs_number in the Patient model, and removes the django M2M ORM implementation between Submission and Patient and replaces with a custom middle table: PatientSubmission
A constraint is added in the `save` method to prevent a submission being saved for the same patient (unique by NHS number) where there is already a record in submission for that audit year. Previous and future audit years can still be saved.
Added tests for the same.


eatyourpeas (commented):
I think this is correct but is a structural change of sorts so would value your review @mbarton if that is ok.

project/npda/models/patientsubmission.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.submission} for {self.patient}"
+
+
+def save(self, *args, **kwargs):
+    # Check for existing submissions for the same patient and audit year
+    if (
+        PatientSubmission.objects.filter(
+            patient=self.patient,
```
mbarton: I think this would need to be a lookup by NHS number rather than patient instance? As they're now not unique by NHS number.

mbarton (approved):

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 36 seconds ago) Please check your changes!

PR #325, opened 2024-10-25
Faster CSV uploads (take 2)

### LINKED ISSUE #147: Validating each postcode on a CSV upload takes multiple seconds

### LINKED ISSUE #148: Getting the GP details takes multiple seconds during CSV upload

mbarton:
Re-spin of https://github.com/rcpch/national-paediatric-diabetes-audit/pull/200

Fixes #148
Fixes #147

Make processing CSV uploads an order of magnitude faster by making all the external API requests in parallel across all rows.

I've done this by making the parent view async along with all the csv_upload functions, and the third party API lookups (`imd_for_postcode`, `validate_postcode` etc). The CSV upload code then creates a Python TaskGroup and submits a task to process each row.

The tasks share an [httpx](https://www.python-httpx.org/) [AsyncClient](https://www.python-httpx.org/async/) so they can co-operatively submit and wait for the responses. For the existing sync calling code on patient create and update I've added wrappers to create a client on demand for the duration of the requests.

I got the approach from this blog post: https://fly.io/django-beats/running-tasks-concurrently-in-django-asynchronous-views/. It's my first time with async Python though so it's very possible I've done something wrong or inefficiently. In particular this approach is naive in proportion to the number of rows in the file. Submitting a CSV with thousands of rows will launch thousands of requests simulataneously which is almost certainly not the behaviour we want. That said, this speeds up our current demo with `dummy_sheet.csv` so much I think it's probably worth merging as is and iterating on. Async code is fiddly by nature so we should not adopt this approach across the board, only where we deem it necessary.

Time to upload `dummy_sheet.csv`:

- *Before*: 12.93 seconds
- *After*: 3.48 seconds! 🚀


project/npda/forms/async_model_form.py
```diff
@@ -0,0 +1,104 @@
...
+
+You can call afull_clean() to run them all asynchronously or use the existing Django
+API calling .errors, .is_valid to run them synchronously. The class keeps track of
+what has been run so regardless of how you call it, they will only run once.
+
+This class mirrors the internals of ModelForm so could break across versions.
+That said, many people override the methods we mirror so they're unlikely to change.
+However you should unit test anything using this to ensure it works across updates.
+"""
+class AsyncModelForm(ModelForm):
```
mbarton: I'm not sure if this abstraction is a good idea, originally I didn't have to pass async_client to each method because I stored it in a contextvar. This was much nicer but I hit issues where the async_client couldn't be used again in a different event loop so the first CSV upload or form submit worked but then subsequent ones failed

project/npda/models/patient.py
```diff
@@ -141,13 +142,31 @@ def age(self, today_date=None):
...
+            try:
+                self.index_of_multiple_deprivation_quantile = await aimd_for_postcode(
+                    self.postcode, async_client
+                )
+            except HTTPError as err:
+                logger.warning(
+                    f"Cannot calculate deprivation score for {self.postcode} {err}"
+                )
+        
+        # HACK: as asave calls save internally and we don't want to call imd_for_postcode twice
```
mbarton: Maybe abandon this hack and just set imd in the calling code (patient form view and csv upload)? That's what I originally did in #200

project/npda/forms/async_model_form.py
mbarton: yeah, canned this as it broke checking NHS number was not null. I don't know why and that demonstrates that this approach was digging too deep overriding Django internals

mbarton:
Overdue on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 15 minutes and 4 seconds ago) What's gone wrong?

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 2 hours, 17 minutes and 48 seconds ago) Please check your changes!

PR #326, opened 2024-10-26
CSV upload improvements

### LINKED ISSUE #315: wrong csv template
The base csv template is published on the NPDA website and should be used for all submissions.

If a user tries to submit data using a csv file with the wrong extension / headings / heading sequence or heading spelling, meaningful rejection message needs to be returned to the user.

### LINKED ISSUE #348: Handle CSV upload missing headers
Some users upload CSV files without headers. We would have to assume they're following the template and parse accordingly

@AmaniKrayemRCPCH to confirm how many users do this

### LINKED ISSUE #349: Don't handle or save CSV upload with additional columns
If they have columns that don't match the template, error out immediately and say what extra columns they need to remove. It's important not to save the CSV file at all in this case as it may contain personal information we are not allowed to store.

NB: if they don't specify any headers at all (#348) then the best we can do is count the columns and error if the count doesn't match

dc2007git:
Fixes #315
Fixes #348 
Fixes #349 

This PR improves CSV uploading:

- Error if we see any unexpected columns
  - This is so we avoid saving data we don't need (which may contain PII)
- Allow uploading without the CSV header
  - We believe some EPR systems export in this format. The best we can do is put the column headers on top and hope for the best.
- Trim whitespace in column names
  - The template CSV published on the RCPCH website has trailing whitespace on one of the columns

Validation errors are currently shown as toast messages after upload.

Thanks @dc2007git for the initial work


dc2007git:
@mbarton @eatyourpeas it looks like between the template csv on the NPDA official website and the dummy invalid csv we're using there's a different of a single whitespace in the template on the thyroid function header. The exact header in the template is 'Observation Date: Thyroid Function ', which if you notice has an extra whitespace character after function that therefore throws an error when compared to an otherwise identical set of headers. If we've already started distributing the template based on the one on the npda website it might be best to leave it at that (and keep the whitespace) to avoid lots of user confusion? What do you both think

anchit-chandran (dismissed):
Looks good, had couple comments:

- Do we need to read a csv file for expected headers? Can we store in just hardcoded constant
- Could we have tests for asserting it returns expected string for working vs not
- I see `difflib` imported but unused. Think would be good for a more meaningful error msg. Particularly because excel files could have lots of weird and wonderful hidden characters that could lead to frustration where people don't know why `NHS number‎ ` is invalid but somehow a secret `\u200B` has snuck in


Re this:
> @mbarton @eatyourpeas it looks like between the template csv on the NPDA official website and the dummy invalid csv we're using there's a different of a single whitespace in the template on the thyroid function header. The exact header in the template is 'Observation Date: Thyroid Function ', which if you notice has an extra whitespace character after function that therefore throws an error when compared to an otherwise identical set of headers. If we've already started distributing the template based on the one on the npda website it might be best to leave it at that (and keep the whitespace) to avoid lots of user confusion? What do you both think

- what if we're less stringent on these tedious things by eg stripping whitespace? Getting awful NHS tech UX flashbacks thinking that a busy consultant is blocked from uploading a file because of random whitespace

dc2007git:
@anchit-chandran great points thanks Anchit. Will remove the difflib import, and completely agree about the whitespace comment. Unless there's a particular reason why (for example something powerDB related on Amani's side of things) I think we can afford to be lenient with things like whitespace, however obviously using the template is ideal.

On that point, the csv file that is read is from the static files (since we provide the template csv) to users, so I thought it would be more consistent to pull the headers from there instead of hard coding them as a constant in the code itself. Let me know what you think is best.

anchit-chandran:
> Will remove the difflib import,

If removing this, can use 2 pointer approach on each `template_header` and `user_header` to check each char is same & also give the exact position&character where they differ to what we expect

> On that point, the csv file that is read is from the static files (since we provide the template csv) to users, so I thought it would be more consistent to pull the headers from there instead of hard coding them as a constant in the code itself. Let me know what you think is best.

Personal feeling is reading in an external file for this is un-needed external dependency where something could go wrong. Think we should hardcode headers as a constant in 1 place, using it in both the function & associated test. For distributing a template file, can add a script that exports those headers into csv (+/- updates wherever users download it from)

project/npda/views/home.py
```diff
@@ -51,20 +56,37 @@ def home(request):
...
+        reader = csv.reader(user_csv.read().decode("utf-8").splitlines())
+        user_headers = next(reader, None)
+
+        template_headers = HEADINGS_LIST
+
+        # Next localise headers so have list and say 'these are the headers affected'
+
+        if user_headers != template_headers:
+            list_diff = [header for header in user_headers if header not in template_headers]
+            error_string = f'CSV headers do not match the expected format/order. Please ensure you are using the template csv provided. Headers affected, or those present in your file but not the template, are: {list_diff}. The headers uploaded must be exactly the same as the template headers, including whitespace, capital letters and ordering.'
```
mbarton: I don't think the order matters? I agree about trimming whitespace on the header names

project/npda/views/home.py
```diff
@@ -51,20 +56,37 @@ def home(request):
     """
     if request.method == "POST":
         form = UploadFileForm(request.POST, request.FILES)
-        file = request.FILES["csv_upload"]
+        user_csv = request.FILES["csv_upload"]
         pz_code = request.session.get("pz_code")
 
-        # summary = csv_summarize(csv_file=file)
+        # You can't read the same file twice without resetting it
+        user_csv.seek(0)
```
mbarton: I wonder if it's easier to pull out reading the CSV file into a dataframe to this level and then pass the dataframe to each function (summarise, check_headers and finally csv_upload?). That way we don't need to remember to seek the file and we guarantee that there's no difference between how Pandas parses the CSV and how the stdlib `csv` module does.

mbarton (commented):
Thanks @dc2007git, I think this is a useful safety check to have.

project/npda/views/home.py
eatyourpeas: good idea: possibly needs some simple validation step here instead then that it is actually a csv with the correct headings and not auntie's investments or the doctor rota uploaded by mistake. If we are not rejecting it, we could store in a dataframe and store the csv in `media` at that point, then run the validation and save steps at that point.

mbarton: @dc2007git are you ok for me to take over this PR and finish it off? There's a few more related issues that came up in the last NPDA catch up which I'd like to fold into this PR: https://github.com/rcpch/national-paediatric-diabetes-audit/issues/348, https://github.com/rcpch/national-paediatric-diabetes-audit/issues/349

dc2007git:
Hi @mbarton sure thing no worries. Am still waiting to get my laptop back from the repair service, sorry it’s taking so long!

project/npda/general_functions/csv_upload.py
```diff
@@ -282,7 +276,7 @@ async def validate_rows_in_parallel(rows_by_patient, async_client):
         return [task.result() for task in tasks]
 
     # Remember the original row number to help users find where the problem was in the CSV
-    dataframe["row_index"] = np.arange(dataframe.shape[0])
+    dataframe = dataframe.assign(row_index=np.arange(dataframe.shape[0]))
```
mbarton, line 370: Fix a Pandas copy on write warning

mbarton (approved):

mbarton:
@anchit-chandran can you have another look at this PR and confirm if you're happy to merge? Thanks!

eatyourpeas:
There is probably some stuff her that is not quite right but for the moment the tests are all passing, it seems to work, so I will take that as a win.

eatyourpeas (approved):
complicated PR. now seems to be working

eatyourpeas (approved):

eatyourpeas (approved):

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @dc2007git and merged by @eatyourpeas 6 minutes and 37 seconds ago) Please check your changes!

PR #340, opened 2024-10-29
Run external validators in parallel

### LINKED ISSUE #330: Set gp_practice_postcode for CSV uploads
It is available in the response from gp_details_for_ods_code. Whilst we don't allow submitting the postcode in the CSV we could add it in the future (#213) so we should ensure in all cases the other field is set.

mbarton:
Speeds up both CSV uploads and creating/updating patients in the UI by issuing all external validation requests (postcode, imd, gp details) in parallel.

Also fixes #330 - gp_practice_postcode is now set for CSV uploads as well as form submissions.

Brings the time taken for CSV upload of `dummy_sheet.csv` down to 1.62 seconds :rocket:

### DIFF
--- a/project/npda/forms/external_patient_validators.py
+++ b/project/npda/forms/external_patient_validators.py
@@ -1,6 +1,7 @@
 import logging
 from dataclasses import dataclass
 
+import asyncio
 from asgiref.sync import async_to_sync
 
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 5 minutes and 28 seconds ago) Please check your changes!

PR #343, opened 2024-10-30
Replace findthatpostcode.uk with postcodes.io

mbarton:
See https://github.com/rcpch/national-paediatric-diabetes-audit/issues/342. findthatpostcode.uk has spikes in latency, replace with something we host and control.

Part 1 of https://github.com/rcpch/rcpch-audit-engine/issues/1096

PR #354, opened 2024-11-08
Use our postcodes.io instance properly

mbarton:
Follow up to #343 and #340.

Make our postcodes.io instance mandatory. **You must update your .env file after this change**.

Actually pass the postcodes.io API key! I didn't notice this stopped working because I was testing with the API key disabled in Azure APIM.

As we don't use a TaskGroup we must re-raise exceptions that occur within our async tasks. Because we already checked for ValidationError, this had the unfortunate effect of silently eating any errors that weren't HTTPError.

I noticed this because I tried to access `response.status` which doesn't exist (it's actually `response.status_code`). But I didn't see anything in the log.

Finally update `dummy_sheet.csv` as it has an invalid postcode. I've left the invalid one in `dummy_sheet_invalid.csv` for demo purposes.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 5 minutes and 25 seconds ago) Please check your changes!

PR #355, opened 2024-11-08
Fix superuser view preference

mbarton:
Fix the default view preference for newly created superusers. It now is correctly set to national where before it ended up being the obselete organisation level (`0`).

This means that you now see the data quality report after uploading in a brand new fresh instance. Before you had to juggle the switcher to get it to show up.

I've refactored the create_superuser function on the user manager to take no additional arguments, which means we can then simplify the calling code to no longer use `**extrafields` and hopefully make bugs like this less likely in the future


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 29 seconds ago) Please check your changes!

PR #363, opened 2024-11-08
questionnaire-upload-division

### LINKED ISSUE #351: uploaders cannot edit data in the questionnaire
To prevent divergence of data between the questionnaire and the upload we need to prevent people who upload from editing and vice versa

eatyourpeas:
### Overview

To prevent users from uploading csvs and then changing the data in the questionnaire this PR adds two new booleans to the session object as well as includes the session object in a new `context_processors` object that is added to settings. This allows the session to be passed automatically into the template without needing to pass it through the context.

The decorator for the two views is also fixed as it was not correctly redirecting. At the same time the home page has been moved to be the dashboard, and the home page instead will be the upload page. the name has not been refactored for the moment.

closes #351


mbarton:
Overdue on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 15 minutes and 1 second ago) What's gone wrong?

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 16 hours, 35 minutes and 22 seconds ago) Please check your changes!

PR #366, opened 2024-11-09
fix-visit-data-type-errors

eatyourpeas:
### Overview
Fixes a few issues in the assigned dtypes on import of the csv to pandas. First the `glucose_monitoring` field was incorrectly cast to `str`. Second the `float32` would not round, so this was changed `float64`. Rounding was introduced for cholesterol, height and weight.
Null values for choices were replaced with None instead of 0
Session is not quite working out if a user has uploaded a csv or not, so raising this as a separate issue.
Small bug fix in the median hba1c as this was trying to calculate medians if some of the values were `None`.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 33 seconds ago) Please check your changes!

PR #394, opened 2024-11-20
audit-year-selection

### LINKED ISSUE #299: CSV uploaders must not be able to edit data in the 'questionnaire'
If users upload .csv files, they should not be able to edit the data subsequently in the user interface as this would cause their data to diverge from the original.

There are two options to this:
1. a workflow that asks lead clinicians their intention on first sign in, and save this decision as a flag which would either switch off access to questionnaire, or given them view only access
2. Allow users to have access to both but once the first patient has been saved by whatever means, this would set the flag constraining them to using this method only going forward.

**Implications**
For either of these options the following things need to happen
- [x] a flag needs adding, possibly to the submission model (since users need only be constrained per submission)
- [x] permissions need updating to reflect the flag value when policing access either to the platform or the csv upload
- [x] we need to decide if users can continue to access the questionnaire if they are spreadsheet users to view their data, or whether we switch this off
- [x] documentation / signposting needs to be made clear to users on login what the options available to them are and that any steps they take are not reversible.

### LINKED ISSUE #384: patient_count is not available in submission_history.html
When viewing the submission history page, `submission.patient_count` is not calculated/is unavailable.

This leaves the entire column empty:
image
_Desktop display - See Patient Number column_

The issue lies in [`submissions.py:49`](https://github.com/rcpch/national-paediatric-diabetes-audit/blob/aa5076b98cf00b90f2225039598e88a8f0a8d014/project/npda/views/submissions.py#L49)

```python
base_queryset.values("submission_date", "audit_year").annotate(
....
``` 

should be

```python
base_queryset = base_queryset.values("submission_date", "audit_year").annotate(
....
``` 
_Note that base_queryset must be mutated_

A seemingly quick fix, but a series of errors follow. Hence, making an issue to be addressed separately.

eatyourpeas:
What started out as a simple fix to the missing patient count in the SubmissionListView turned into a new feature to include an audit year selector

This PR places a new select in the navbar alongside the `view_preference` selector. To implement this it adds:
- `audit_year_select` partial
- `audit-year` url for the htmx `POST` request on change of the select
- store the `selected_audit_year` in session
- a new custom htmx trigger to requery all submissions, patients and visits on update of the session
- 2 new permissions mixins - one to prevent access to create/update/delete routes if `can_complete_questionnaire` session is `False` and another if users are attempting to change, create or delete records in previous audit years
- a new test class for the routes (at least the create routes)

### Overview
Please describe the purpose of this PR here.

### Related Issues
closes #384

### Mentions
@reecehill the new select does not quite look as it should. Could you review and make it works more like the `view_preference` one that you made? This audit_year_select  need not have an apply/remove button but would be good if the filter might be visible in the tables like the `view_preference` one


eatyourpeas:
image

Patient number now fixed - no patient number show for inactive submissions since the data has been deleted.
The `audit_year_select` though is not quite right

reecehill:
@eatyourpeas  This the kind of thing you had in mind?

Note that I've condensed the filters box as it was beginning to present duplicate info, e.g., the current audit year would be displayed twice (once in grey, once in dropdown). The update has also improved mobile responsiveness as a result. 

image

image

Ready for your review.

reecehill (changes requested):
@eatyourpeas Please go ahead and merge once you're happy.

reecehill (approved):

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @eatyourpeas and merged by @reecehill 9 minutes and 43 seconds ago) Please check your changes!

PR #395, opened 2024-11-23
dont-lose-your-head

### LINKED ISSUE #391: Reject CSV uploads missing header
@AmaniKrayemRCPCH confirmed we don't need to support this: https://github.com/rcpch/national-paediatric-diabetes-audit/issues/348#issuecomment-2488221842

This issue is to remove the code that prepends the header if it's missing and to reject the CSV if the headers are missing

eatyourpeas:
Fixes #391

### Overview
Uploading a CSV without a header is now seen as a fatal error (as it could be the doctor rota which they have uploaded, not the audit data) and the file is rejected. 

Now  a `ValueError` is raised and this is trapped in the home page and passed back to the user in an error message and a redirect. 

The test for this (reassuringly failed after the revert was made) now captures the error and checks not records have been saved

### Related Issues
Closes #391

### DIFF
--- a/project/npda/general_functions/csv/csv_parse.py
+++ b/project/npda/general_functions/csv/csv_parse.py
@@ -3,6 +3,9 @@ from dataclasses import dataclass
 import logging
 import re
 
+# Django imports
+from django.core.exceptions import ValidationError
+
 # Third-party imports
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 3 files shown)


eatyourpeas:
image

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 10 minutes and 47 seconds ago) Please check your changes!

PR #397, opened 2024-11-26
Use api.rcpch.ac.uk for NHS organisations and Digital Growth Charts

mbarton:
Force the use of API keys and our API management platform for the NHS organisations and Digital Growth Charts API calls. This is so we can effectively limit access per application if something goes wrong.

@anchit-chandran @eatyourpeas I have removed all the calls to the NHS organisations API that are not currently used, including one from `misc_py_shell_code.py`. Please take a look and if I've removed something we still need let me know.


project/npda/tests/factories/npda_user_factory.py
```diff
@@ -73,12 +71,6 @@ def organisation_employers(self, create, extracted, **kwargs):
         else:
             # If pz_codes are provided, create OrganisationEmployer for each pz_code
             for pz_code in extracted:
-                pdu_data = get_single_pdu_from_pz_code(pz_number=pz_code)
```
mbarton, line 76: This wasn't used anywhere and this is test code so I don't think we need to validate the input?

project/npda/general_functions/rcpch_nhs_organisations.py
```diff
@@ -15,205 +15,29 @@
...
-    url = f"{settings.RCPCH_NHS_ORGANISATIONS_API_URL}/paediatric_diabetes_units/parent"
-    ERROR_RESPONSE = [("999", "An error occurred while fetching NHS organisations.")]
-
-    try:
-        response = requests.get(url=url, timeout=10)  # times out after 10 seconds
-        response.raise_for_status()
-
-        return response.json()
-
-    except HTTPError as e:
```
mbarton, line 161: This code was called from a seeding function via `manage.py` so I don't think there's much value in trapping errors, just let them bubble up and crash the command

project/npda/tests/factories/npda_user_factory.py
anchit-chandran: I think this pdu_data was being used back when we had more of a mix with ods codes; can remove!

I ?think I was making that assertion because it was an external api call (which should anyway be mocked for these tests)

anchit-chandran:
FYI `misc_shell_code.py` could probably be removed, just had it so you could easily copy paste factory code to quickly seed objects from the manage.py shell. However that's more "rough" work, so need not be in main codebase

eatyourpeas (approved):

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 9 minutes and 40 seconds ago) Please check your changes!

PR #405, opened 2024-11-28
move-centile-calculation-into-form

### LINKED ISSUE #401: dGC API not called when visits added using questionnaire
I think this is an oversight rather than deliberate @eatyourpeas?

eatyourpeas:
Fixes #401

### Centile calculations have moved
Previously being called in the async method, the API calls have moved now into the form instance itself. This means that the centile is called and persisted whether the heights and weights are entered in the questionnaire or in the csv upload


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 11 minutes and 52 seconds ago) Please check your changes!

PR #416, opened 2024-12-04
Map!

eatyourpeas:
### Overview
Adds a scattermap using plotly
Adss the mapbox/carto API key to the config first
Filters children against audit year and PDU and aggregates distances from the the PDU.
Passes these to the dashboard template

This involves changes to the model by adding geo location to the model

Next step is to save the geolocation from postcode during validation steps but that is async so will add separately. Currently it is calculated on landng on the page


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 9 minutes and 46 seconds ago) Please check your changes!

PR #388, opened 2024-11-19
centiles-for-measurements

eatyourpeas:
### Overview
Integrates the digital growth charts API to calculate centiles and SDS where heights and weights are provided
Reports these values to the user in the questionnaire

### Code changes
New file `dgc_centile_calculations` which makes the API calls. Note that credentials need to be added to the .env (see settings.py)
`csv_upload` makes the api calls within the async work flow prior to saving the visit
New fields for height/weight/bmi centiles and SDS are added to the Visit model so generate a new migration
These fields are added to the form, but have to set again to None on cleaning otherwise are cast to empty strings which then fails validation
The results are then pulled into the template and have a new template tag to help with this

### Documentation changes (done or required as a result of this PR)
Please describe any changes to documentation here.

### Related Issues
List any issues related to this PR here.

### Mentions
@mentions of the person or team responsible for reviewing proposed changes.

PR #418, opened 2024-12-05
Call Digital Growth Charts API asynchronously

### LINKED ISSUE #399: Pass through warnings from dGC API
The Digital Growth Charts API returns warnings if values for height and weight exceed expected ranges (by standard deviation).

- https://github.com/rcpch/rcpchgrowth-python/issues/32
- https://github.com/rcpch/rcpchgrowth-python/pull/45

At the moment these are logged to the console but I'd argue they should be data validation errors. They have been calibrated against datasets of real measurements so false positives are very unlikely. For CSV uploads anyway we'd still store the underlying data anyway.

### LINKED ISSUE #415: CSV upload calls dGC in sequence
#405 regressed CSV upload performance as the dGC calls are no longer made asynchronously. This means it can't make progress until the API response is back.

This flame graph is from uploading a 1600 row CSV file (100 patients).

Screenshot 2024-12-03 13 26 36

mbarton:
Fixes #399 
Fixes #415 

Follow on from #388 and #405, moving the calls to the Digital Growth Charts API to be made asynchronously during CSV upload.

Uploading a CSV of 100 patients with 16 visits each (1600 rows):

- *Before*: 66 seconds
- *After*: 21.5 seconds

Also pass through warnings from the API as validation errors. This catches height and weight measurements that are so extreme they are wrong (see https://github.com/rcpch/rcpchgrowth-python/issues/32 for the methodology)


project/npda/forms/visit_form.py
```diff
@@ -61,15 +58,7 @@ class Meta:
             "hospital_discharge_date",
             "hospital_admission_reason",
             "dka_additional_therapies",
-            "hospital_admission_other",
-            # calculated fields
-            "height_centile",
```
mbarton, line 66: As these are calculated fields it should not be possible to submit them through the form

project/npda/forms/visit_form.py
```diff
@@ -740,114 +708,59 @@ def clean_hospital_discharge_date(self):
...
-                    cleaned_data["bmi_centile"] = round_to_one_decimal_place(centile)
-                    cleaned_data["bmi_sds"] = round_to_one_decimal_place(sds)
-                except Exception as e:
-                    cleaned_data["bmi_centile"] = None
-                    cleaned_data["bmi_sds"] = None
-                    # we are not raising a validation error here as sds and centile are not required fields
-                    pass
+            cleaned_data["weight"] = weight = round_to_one_decimal_place(weight)
+
+        if not getattr(self, "async_validation_results", None):
```
mbarton, line 748: This is the same trick we already use for the patient form, allows us to run things asynchronously for the CSV upload but synchronously for the questionnaire form submission

project/npda/management/commands/upload_csv.py
```diff
@@ -0,0 +1,46 @@
+from asgiref.sync import async_to_sync
```
mbarton, line 1: I wrote this for performance profiling as you can put cProfile in front of it easily (https://stackoverflow.com/questions/582336/how-do-i-profile-a-python-script).

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 32 seconds ago) Please check your changes!

PR #431, opened 2024-12-14
Eatyourpeas/issue422

### LINKED ISSUE #422: Map render in htmx partial view crashing on dashboard
*Merging #421 into live so this can be worked on in own branch so no worries about disruption*

The map render logic has been abstracted into its own partial htmx view `get_map_chart_partial` in `dashboard.py`. 

The `dashboard` view will pass `get_map_chart_partial` the pdu.pk and selected_audit_year (inside the `imd_map_card.html` -> **NOTE: CURRENTLY COMMENTED OUT as crashing, so just need to comment back in when working on this**. This view performs the map logic code and finally returns a `go.Figure` that is converted to HTML and should then be passed back into template.

The setup logic has been slightly tweaked to do a query to get the pdu from the pk (as couldn't serialise a PDU object via GEt request). Also updated `generate_distance_from_organisation_scatterplot_figure` to return a `Figure` so we can convert straight to html

Using debugger I noticed these 2 parts are the slowest:

Image

Afaik it does successfully return the figure at the end (code should be working) but something crashes, including crashing the debugger itself aha so struggling to fix

I think important fixes would include:

- optimising load through the obvious stuff you already mentioned, radius, lower detail in map render etc
- seeing if that in itself fixes the issue
- try to avoid setting an explicit height / width anywhere, hoping to rely on plotly's auto dimensioning to fit itself within the template properly

eatyourpeas:
### Overview
Implements maps as an ansynchronous htmx get request on dashboard load

This adds a map (using a MapBox base map which requires an API key), centred on the PDU, with a scatterplot of all the children associated with that PDU, with distance to the PDU and the NPDA ID in a tooltip on hover.
The base map has overlaid the local authority boundaries in black, and the LSOA boundaries filled with dark pink to dark blue based on deprivation rank (using the 2019 IMD data)

Because the geojson files with the geometry are so large, these are added in the constants and pulled into a geopandas dataframe on landing on the PDU. 

A new endpoint on the RCPCH NHS Organisations API was added to return all local authorities within a radius of the PDU lead centre longitude and latitude (I have picked 10 km but this could be changed). The LSOA and LAD local geopandas dataframes are filtered to these to speed up render time.

There are some issues that will need addressing to optimise this process:
1. The patients currently have their long and lat calculated from postcode with an API call and persisted in the model in the dashboard view - ideally this should happen at the time the patient is save to the database. This will need adding therefore to the async functions and removing once that is done. I will raise a separate issue for that
2. Amani said that quite a few Welsh patients actually live in England, so with the current implementation, they will be plotted on the white basemap without English LSOAs, LADs and deprivation data.
3. It was agreed at recent sprint that the colour scales for IMD in England and Wales should be different, since they are based on different reference data from the same year (English IMD 2019, vs WIMD 2019) - I will raise a separate issue for this as it is related to 2.

### Code changes
- ONS Data files added as geojson to an new English IMD 2019 folder in `constants`. This actually already is inappropriately named and should be refactored to IMD and boundaries or similar
- the `generate_distance_from_organisation_scatterplot_figure` function in `map.py` updates the colorscale, loads in the geometry for LADs and LSOAs as well as the English or Welsh IMD ranks/deciles (already merged as separate files). It then makes a new API call to get all local authorities within 10km of the PDU lead centre and generates geopandas dataframes from the loaded files with the unwanted LADs filtered out. It then generates a map with Choropleths for LADS and LSOAs, and 2 scatterplots, one for patients, one for the PDU lead centre. This is returned to the dashboard view and in turn returned to the template.
- `rcpchnhsorganisations.py` - includes a new API call to the new endpoint accepting `long`, `lat` and `radius` (m) parameters.

I will add this in subsequent commits

### Related Issues
List any issues related to this PR here.

### Mentions
will close #422

Additional issues to raise:
1. handling welsh patients in england: implications for mapping (colour scales between the two nations, plotting English and Welsh data together)
2. saving geocoordinates of patients on save


eatyourpeas:
image

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 26 seconds ago) Please check your changes!

PR #438, opened 2024-12-15
async-patient-location-save

### LINKED ISSUE #417: location data should update Patient model asynchronously
The Patient model now has geos fields to store patient location based on postcode.
To convert from postcode to long/lat an API call to postcodes.io is needed first to retrieve the long,lat and these can then be converted to a gis Point. 
These new fields need at the same time to be excluded from the patient form

eatyourpeas:
### Overview
Location save is added to the async validation functions when a patient is added to NPDA, either by csv upload or through the questionnaire.

This builds on the elegant work of @mbarton. It adds a new async function that uses the postcodes.io service to get longitude and latitude against the patient postcode. It uses these to generate two geodjango Point objects (one for the bng or british national grid which uses northings and eastings, the other for wgs84 which uses the crs 4326 which is standard for mapping). A small optimisation is that the imd and location async functions are now ignored if the postcode fails validation. This limits the noise in the logs.

### Code changes
- `external_patient_validators`: adds a new `_location_for_postcode` method, and conditionally calls this or `_imd_for_postcode` if postcode is valid. Location is returned as a tuple of `location_wgs84, location_bng`
- `patient_form`: adds the new location fields and updates them on save
- adds the location fields to the `csv_upload` 
- removes the creation of location fields from `map.py` which had been there as a temporizing measure
- makes the `location_for_postcode` function async

closes #417

### DIFF
--- a/project/npda/forms/external_patient_validators.py
+++ b/project/npda/forms/external_patient_validators.py
@@ -14,6 +14,7 @@ from ..general_functions import (
     validate_postcode,
     imd_for_postcode,
     calculate_centiles_z_scores,
+    location_for_postcode,
 )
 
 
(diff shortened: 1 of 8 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 28 seconds ago) Please check your changes!

PR #443, opened 2024-12-17
Broken maps

### LINKED ISSUE #442: broken maps
In the process of updating the RCPCH NHS Organisations API to ensure border coordinates for mapping returned were geojson, the response objects in NPDA now need remapping.

eatyourpeas:
The maps broke after an update to the RCPCH NHS Organisations API to add geojson geometry. 

This also changes the LAD boundaries bundled into NPDA to the 2011 boundaries and swap out the 2024 boundaries. This matches the IMD data which is 2019 but laid into the 2011 boundaries. It is more appropriate I think to use this even if the boundaries are not completely in step with what is in place today.

### DIFF
--- a/project/npda/general_functions/map.py
+++ b/project/npda/general_functions/map.py
@@ -237,7 +237,6 @@ def generate_distance_from_organisation_scatterplot_figure(
     )
     lad_gdf = gpd.GeoDataFrame.from_file(file_path)
     # filter out unwanted LADs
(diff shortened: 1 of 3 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 33 seconds ago) Please check your changes!

PR #444, opened 2024-12-20
Detailed-lsoa

eatyourpeas:
### Overview
The LSOA boundaries currently are quick to load but look ugly
In this PR the boundaries are updated to be generalised clipped, replacing the supergeneralized boundaries. The files are bigger (so are slower to load) but look nicer. Let's try this, we can always go back.

In addition, the files are >50MB so trigger the github warning advising us to use github LFS. The files have not been thrown out but if we are going to keep them we shold probably think about that


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 32 seconds ago) Please check your changes!

PR #445, opened 2024-12-21
updates map call without geojson

eatyourpeas:
### Overview
Quick fix to roll back the geojson call


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 35 seconds ago) Please check your changes!

PR #473, opened 2025-01-11
Improve patient table on smaller screens

mbarton:
### Before

Screenshot from 2025-01-11 16-24-11

### After

Screenshot from 2025-01-11 16-23-57

### Before
Screenshot from 2025-01-11 16-22-59

### After
Screenshot from 2025-01-11 16-22-36

- Remove some columns to make space
  - NPDA patient ID
  - Date uploaded
  - Audit year
- Correctly call the IMD column quintile not decile
- Change the text to black not grey to improve readability
- Replace all SVG icons with font awesome to reduce the amount of code in the template
- Fix the visit tooltip on error

Fixes #468, although the link still goes off screen at lower sizes.

PR #474, opened 2025-01-11
Save cleaned fields during CSV upload even if other fields are invalid

mbarton:
I noticed testing #473 with `dummy_sheet.csv` that the NHS number had spaces in it if the CSV row had an error in one of the patient columns but otherwise it didn't.

CSV upload is implemented by filling in the same Django forms as manual input using the questionnaire. This was a bug where we only saved the cleaned data if the entire form were valid. So if you had a validation error in one of the columns that maps to the patient form, none of the other cleaned values would be saved (and the same for visit form).

This has real world implications as the normalised NHS number would not be saved, potentially meaning duplicates would not be detected during upload.

The fix was simply to take all the fields from cleaned data and then fill in any extra ones from data.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 8 minutes and 41 seconds ago) Please check your changes!

PR #477, opened 2025-01-12
Show patient NHS number in visit form

mbarton:
We previously showed the database ID which is not useful for cross referencing with other systems. Since NPDA doesn't collect names this is the best identifier we have.

**Before**

Screenshot from 2025-01-12 16-26-48

**After**

Screenshot from 2025-01-12 16-26-05


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 34 seconds ago) Please check your changes!

Issue #481, opened 2025-01-13
Adding new audit years


mbarton:
We need to decide how new audit years are added, opened and closed for submission on the platform.

We can hard-code them or have a model managed through the Django admin. I think my preference is for the latter as it allows audit staff to fully manage any grace periods needed. But we are able to make code changes pretty quickly so it might be too much engineering for NPDA and one for the next audit platform?

PR #482, opened 2025-01-13
Fix audit year when logging in Jan to April

### LINKED ISSUE #460: Default selected audit year not just `datetime.now().year`
@eatyourpeas was looking to update this myself now but unsure if it's set / used in multiple places, and unsure if I'm right, so wanted to confirm with you

Currently `session.py::65`, in return of `create_session_object()`, we set the default `selected_audit_year` as:

```py
"selected_audit_year": datetime.now().year,
```

But I think that because audit periods are laid out like:

```
        Audit Period	Audit period start	Audit period end
        2024 - 2025	        1-Apr-24	        31-Mar-25
        2025 - 2026	        1-Apr-25	        31-Mar-26
        2026 - 2027	        1-Apr-26	        31-Mar-27
```

the default audit year for e.g. `4 Jan 2025` ISNT `2025` but is `2024`?

We have a `get_audit_period_for_date()` fn inside `general_functions/audit_period.py` that gets audit period for given date so simplest change could be like:

```
get_audit_period_for_date(datetime.now())[0].year 
```

mbarton:
Fixes #460. Audit years run April -> March so the current audit year (as of 13 Jan 2025) is 2024 not 2025.

I've raised #481 to track how we want to add new audit years.

### DIFF
--- a/project/npda/general_functions/audit_period.py
+++ b/project/npda/general_functions/audit_period.py
@@ -3,6 +3,15 @@ from dateutil.relativedelta import relativedelta
 
 
 # TODO MRB: How will we add new audit years (https://github.com/rcpch/national-paediatric-diabetes-audit/issues/481)
+SUPPORTED_AUDIT_YEARS = [
+    # submitted on the old platform, re-uploaded to test this one
+    2023,
+    2024,
+    # first year submitted on this platform
+    2025,
+    2026
+]
+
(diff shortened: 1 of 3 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 6 seconds ago) Please check your changes!

PR #484, opened 2025-01-14
Fix incorrect usages of current year as audit year

### LINKED ISSUE #460: Default selected audit year not just `datetime.now().year`

mbarton:
Fixes #460 (again!)

The audit year runs April -> March. This PR fixes the remaining references to the current year when:

- Uploading a new submission via CSV
- Creating the first patient in a new submission using the questionnaire
- Setting the audit year switcher when logging in

It also removes unused code that referenced the current year:

- `csv_summarize`
- `@only_update_current_audit_year`


mbarton:
The test failure is happening because I moved setting `submission.errors` into the `csv_upload` function.

Unfortunately the default pytest Django setup doesn't let you make any further database calls after something that would cause a transaction to rollback (https://stackoverflow.com/questions/21458387/transactionmanagementerror-you-cant-execute-queries-until-the-end-of-the-atom).

I fixed that for the database calls in the test assertions by adding transaction.atomic: https://github.com/rcpch/national-paediatric-diabetes-audit/commit/cc37e637dd4ce1d3d8a851ff4526ebda92345a0f. That won't work here though without the call being within the `csv_upload` function itself and in our current data model we don't want to add all the CSV rows as a single transaction (instead we do it within a separate `Submission`).

If I can't think of anything better I'll just put the code back to how it was, setting `submission.errors` in the view.

mbarton:
[Copied from commit message about fixing the other tests for missing mandatory fields]

The first one worked because we ran the seed_users fixture which spawns the appropriate PDU for the user if it doesn't exist already. The problem is that was a session fixture, so it didn't run it again on the next test run.

No biggie for normal tests as the database is rolled back to the state after the fixture ran, including the PDU. But for the transactional tests that doesn't work as it truncates all the tables after each test.

So my solution is to run the fixtures before each test (function fixtures). It's slow but still only a 40 second test run (was 27 seconds before) so not a big deal and I think it will be useful to have all the missing mandatory fields covered.

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 41 seconds ago) Please check your changes!

PR #490, opened 2025-01-16
Fix sorting by IMD and sort icon direction

mbarton:
The sort icons are now greyed out until you click them to activate sorting by that column ascending. Click again for descending. Previously the icons were pointing in the wrong direction after you clicked them.

Unfortunately this breaks the cool rotating animation, not sure how to put that back as it was the client side toggle that led to them pointing in the wrong direction.

[Screencast from 16-01-25 15:19:16.webm](https://github.com/user-attachments/assets/d32a435a-9562-4291-957d-7770a0bdbef9)

*Unfortunately the mouse pointer is missing from the video above. You trigger the sorting by clicking on the icon!*


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 40 seconds ago) Please check your changes!

PR #494, opened 2025-01-20
Patient list UX improvements

mbarton:
Screenshot from 2025-01-20 11-39-19

- Put errors first, unless you have sorted by a different field
- Put the page controls at both the top and the bottom of the page
- Fix the total count in the page controls
- Fix the divider between valid and invalid to show the correct count for the current page
- Fix the divider to only show on pages with a mix of valid and invalid rows

### DIFF
--- a/project/npda/views/patient.py
+++ b/project/npda/views/patient.py
@@ -108,28 +108,7 @@ class PatientListView(
         if sort_by:
             patient_queryset = patient_queryset.order_by(sort_by)
         else:
-            patient_queryset = patient_queryset.order_by("is_valid", "-visit_error_count")
-
-        # Add extra fields to the patient that we can't add to the query. This is ok because the queryset will be max the page size.
-        seen_error = False
-        seen_valid = False
-
-        for patient in patient_queryset:
-            # Signpost the latest quarter
-            if patient.most_recent_visit_date is not None:
-                patient.latest_quarter = retrieve_quarter_for_date(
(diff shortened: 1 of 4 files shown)


project/npda/views/patient.py
```diff
@@ -162,32 +129,43 @@ def get_context_data(self, **kwargs):
...
+        context["error_count_in_page"] = error_count_in_page
+        context["valid_count_in_page"] = valid_count_in_page
+
         return context
 
     def get(self, request, *args: str, **kwargs) -> HttpResponse:
         response = super().get(request, *args, **kwargs)
+        
         if request.htmx:
-            # filter the patients to only those in the same organisation as the user
```
mbarton, line 170: Everything still seems to work without this HTMX specific override. @eatyourpeas can you remember why we needed to have this special case?

eatyourpeas: for sure could be removed. It ensures returning only the table partial template if it is an htmx request, rather than returning the whole page which is what would happen with a page reload. It is probable that it doesn't get called currently except from that htmx get request which i think only happens once the main page has loaded.

mbarton: Discussed in person and we are happy this can be removed

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 38 seconds ago) Please check your changes!

PR #514, opened 2025-01-25
jersify

### LINKED ISSUE #509: The other half of double Jersey
We need to implement Jersey

eatyourpeas:
### Overview
Adds Jersey to NPDA

Complicated because Jersey does not use NHS Number but instead uses Unique Reference Number.

Since NHS Number is not unique, URN can't be unique either, so it is implemented the same as NHS Number and where Jersey is used, the heading in the CSV is replaced with URN. This starts with the template CSV and tracks all the way through the application.
The forms and validations conditionally use URN depending on what is in session. URN is used if the user has selected Jersey in the selector.

This generates two new migrations.

Outstanding work includes updating the Patient factory and seed csv functions to accommodate the changes elsewhere


eatyourpeas:
Probably should sense check @mbarton first before merging in.

A broader summary of the changes made by this PR

1. *Unique Reference Number*
- Jersey does not have NHS Numbers, instead it has Unique Reference Numbers. There is no real validation of these, beyond the fact that they are numbers. I have put an absolute length of 50 digits, but could shorten this if needed.
In the process of adding URNs I have had to make NHS Number allow None values and have therefore added a clean method to the model to prevent an instance of the model saving with no identifier beyond the primary key. Be aware that this `clean` method will not get called on every save - it will only be run if `full_clean()` is called on the instance before saving. 
- Because there is no NHS Number, the CSV file had to change. It is identical to the NHS file with the exception of the Unique Reference Number column header. Conditionals are introduced that return the right list of column headers depending on the PZ Code in the session.
- A new change is that if the NHS number and Unique Reference number are both missing for any fields, the whole CSV upload is rejected. This happens in `csv_parse`.
- The csv creation functions have been updated to include a `--pz_code` parameter to allow for Jersey (PZ248). In doing so the `--build` attribute is now mandatory. 

You will need a Jersey csv:
```console
python manage.py create_csv \
        --pts=5 \
        --visits="CDCD DHPC ACDC CDCD" \
        --hb_target=T \
        --age_range=11_15 \
        --pz_code="PZ248" \
        --build
```
2. *Maps*
- The Jersey maps fortunately were already in. There are no deprivation data and no LSOAs so shape files to overlay on the base map.
3. *Templates*
- These have been updated to render URN instead of NHS Number where needed. There is now inconsistency if you change the PDU selected midway through completing a form in the questionnaire or if viewing a submission or patient list - the tables update but the headers don't because the `hx-get` only applies to the table partials, not the template within which they are embedded. I can come up with a solution for that, but probably we need to disable the selector while people are actually in a form.
4. *Tests*
- there are no tests currently for Jersey, nor is there any documentation about it.

eatyourpeas (commented):
Took a little while to bring this PR up to date with live

It now seems to work as expected but I have a couple of questions to the NPDA dev brain before merging in.

Previously NHS number was a mandatory field, but now it has been made `null=True` which has a knock on effect for the way it is cleaned in the form. There is a new UniqueReferenceNumberField in the form and validation methods in the form and the model, which has got me to thinking:

Are we generally happy with the principle that a record should not be saved under any circumstances if the the NHS Number or Unique Reference Number are invalid or neither is supplied? If so, that is how it currently is. There are tests at the moment for NHS Number not supplied but not for URN.

mbarton:
> Are we generally happy with the principle that a record should not be saved under any circumstances if the the NHS Number or Unique Reference Number are invalid or neither is supplied? If so, that is how it currently is. There are tests at the moment for NHS Number not supplied but not for URN.

Yes, I think you either have one or the other and not both.

Thinking ahead to the generic audit platform I'm wondering if we should have an `external_id` field which covers both scenarios and move NHS number validation to the form or Pydantic model? Probably too big a change for NPDA now.

mbarton:
> There is now inconsistency if you change the PDU selected midway through completing a form in the questionnaire or if viewing a submission or patient list - the tables update but the headers don't because the hx-get only applies to the table partials, not the template within which they are embedded. I can come up with a solution for that, but probably we need to disable the selector while people are actually in a form.

I copped out in #531 and made changing the switcher a full page reload, which should fix this too

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 43 seconds ago) Please check your changes!

PR #519, opened 2025-01-27
[Auto Logout Middleware] Adds session logout based on idle time

### LINKED ISSUE #491: Auto logout
Check how long our sessions last. We should automatically log users out after a certain period of time. E12 does 30 minutes which I find too aggressive but the timeout should definitely be less than a day.

anchit-chandran:

As related to security, wanted review from both to ensure it's safe. Uses https://pypi.org/project/django-auto-logout/

### Overview

Adds middleware that tracks authenticated users' last activity, auto-logging out after env var `AUTO_LOGOUT_IDLE_TIME_SECONDS` seconds

Adds test for this too.

### Code changes

- test creation, settings reading in env var with warning if missing, fix bug in `login_and_otp_required`

New requirements:
- freezegun (for ease of testing with times)
- `django-auto-logout`
- isort

# Related issues
#491 
#542


project/settings.py
```diff
@@ -190,6 +186,11 @@
 SESSION_COOKIE_HTTPONLY = True  # cannot access session cookie on client-side using JS
 SESSION_EXPIRE_AT_BROWSER_CLOSE = True  # session expires on browser close
 
+# Auto-logout
+AUTO_LOGOUT_IDLE_TIME_SECONDS = os.environ.get(
```
anchit-chandran: Can be adjusted @mbarton to a more suitable time

mbarton: This didn't work when configured from the env file.

Capture

I have pushed a fix

project/npda/templates/dashboard/components/cards/ethnicity_card.html
```diff
@@ -4,8 +4,8 @@
 {% endblock card_title%}
 {% block card_body %}
 <div
-hx-get="{% url 'get_waffle_chart_partial' %}"
-hx-vals='{{ charts.pt_ethnicity_value_counts_pct.data }}'
+hx-get="{% url 'get_treemap_chart_partial' %}"
```
mbarton: @anchit-chandran please can you split out the treemap changes to a separate PR with it's own description. Thanks!

mbarton (changes requested):

mbarton:
Thanks @anchit-chandran, I've tested this and it works for me but I'm concerned about the UX.

I do think there's value in us having a Javascript snippet that sends people to the login page before they try and do an action ([django-auto-logout](https://pypi.org/project/django-auto-logout/) provides this in E12). It would be frustrating to try and do something, maybe even fill in part of a form only to then be logged out.

project/settings.py
anchit-chandran: Thanks! Good catch. I'd misspelt my env var to `AUTO_LOGOUT_IDLE_TIME` so was resorting to default; just pushed a commit warning if that's happening

project/npda/templates/dashboard/components/cards/ethnicity_card.html
anchit-chandran: done

anchit-chandran:
Refactoring to show msg as:

image

mbarton (approved):
Looks good and I've tested it locally. Thanks @anchit-chandran!

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @anchit-chandran and merged by @mbarton 6 minutes and 39 seconds ago) Please check your changes!

PR #522, opened 2025-01-28
Show error in tab when submitting using the questionnaire

### LINKED ISSUE #507: Validation error is hidden after submission in second visit tab
Create a visit with an invalid ACR measurement (eg 80).

The form returns an error but you can't see it because it's sent you back to the Routine Measurements tab.

~~There's also maybe a bug with the tab level error icons from #499 - the Annual Review tab isn't showing the error icon~~ fixed in https://github.com/rcpch/national-paediatric-diabetes-audit/pull/508

mbarton:
Fix #507 by automatically loading the first tab with an error in it after submitting the form.

This is a big refactor where I've finally bit the bullet and removed all the template filters with a nested dict describing the three tabs and associated categories.


project/npda/general_functions/visit_categories.py
```diff
@@ -1,28 +1,85 @@
...
+            for field, field_errors in category["errors"].items():
+                errors[field] = field_errors
+
+        tab = {
+            "name": tab_name,
+            "categories": categories,
+            "errors": errors
+        }
+
+        # Show the first tab with errors
```
mbarton, line 74: The javascript helper also does this and scrolls to the right place but can't hurt to have the HTML render as expected anyway

project/npda/templates/npda/visit_form.html
```diff
@@ -61,4 +60,36 @@
       </form>
     </div>
   </div>
+  <script>
```
mbarton, line 63: This script was in `visit_form_tab` before so was run 3 times. I've moved it here to run once.

The links no longer have the `tab` parameter and we look up the right tab in the script from the anchor

project/npda/templates/partials/visit_form_tab.html
```diff
@@ -4,114 +4,91 @@
...
               </span>
             {% endif %}
-            {{ field_category }}
+            {{ category.name }}
           </a>
         </li>
       {% endfor %}
     </ul>
   </div>
-  {% for field_category in categories|split_by_comma %}
```
mbarton, line 28: This diff looks horrendous but it's the same code with one less level of nesting as we can now remove the `with` for background colour

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 8 minutes and 5 seconds ago) Please check your changes!

PR #524, opened 2025-01-29
Fix dGC height, weight and BMI fields for CSV upload 

### LINKED ISSUE #523: BMI and centiles not working for CSV uploads
They do appear for manual entry through the questionnaire

mbarton:
Fixes #523 

The code to save these fields was in `VisitForm.save()`. This wasn't being called during CSV upload.

I had previously tried this with `PatientForm.save()` but hit two road blocks:

- There was no asynchronous version to call
- The `ModelForm` base implementation throws a `ValueError` if validation has failed
  - For CSV uploads we still want to save the instance anyway and errors are stored on a separate field

I've decided the best thing to do is just to call these methods, as otherwise it's likely we'll forget to update both places where derived fields need to be saved. Our overrides of `save` don't call `super().save` but that's not a problem in current versions of Django. The risk is that future versions will add functionality in there that we'll miss out on. But for CSV uploads at least the functionality is heavily unit tested at a higher level.

I've tested a 1600 row CSV (100 patients, 16 visits each) and it takes the same amount of time to upload as before, so the new `sync_to_async` calls don't regress performance in any practical sense.

### DIFF
--- a/project/npda/forms/patient_form.py
+++ b/project/npda/forms/patient_form.py
@@ -179,17 +179,20 @@ class PatientForm(forms.ModelForm):
         ]:
             self.handle_async_validation_result(key)
 
-    def save(self, commit=True):
-        instance = super().save(commit=False)
+    def save(self):
+        # We deliberately don't call super.save here as it throws ValueError on validation errors
+        # and for CSV uploads we don't want that to stop us. As of Django 5.1.5 it doesn't do anything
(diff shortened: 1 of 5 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 8 minutes and 46 seconds ago) Please check your changes!

PR #516, opened 2025-01-26
implement distance to PDU

eatyourpeas:
### Overview

Implements distance from PDU in the patientlist table

PR #527, opened 2025-01-29
Fix patient list overflowing on smaller screens

mbarton:
Adding the distance to the PDU (#516) overflows the patient list on smaller screens so you can't click the visits link.

This PR gains back some space by:

- Removing the Ethnicity column (need to check with the team this is OK)
- Renaming `Deprivation Quintile` column to `IMD` and adding a tooltip describing what it is
- Reducing the padding around the table from 10 to 1

The table is now good to display without overflow down to about 1200px wide. I think this will include most desktops and laptops people are likely to use NPDA on but there is an open question how we could make the "Visits" link accessible even if the table overflows on something smaller (https://github.com/rcpch/national-paediatric-diabetes-audit/issues/528)

**Before**

Capture

**After**
Capture

### DIFF
--- a/project/npda/templates/partials/patient_table.html
+++ b/project/npda/templates/partials/patient_table.html
@@ -13,22 +13,23 @@
               NHS Number
             </span>
           </th>
-          <th scope="col" class="px-2 py-3 text-center">Sex</th>
           <th scope="col" class="px-2 py-3 text-center">Date of Birth</th>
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 41 seconds ago) Please check your changes!

Issue #538, opened 2025-01-31
Performance testing


mbarton:
Upload big CSV file (10,000 rows)
Upload big files across multiple PDUs (10?)
Test dashboard perf (#453)
Test patient list perf (seems to take a while to load initially)

anchit-chandran:
Dashboard performance test init #634

Issue #498, opened 2025-01-21
Harden access to the config share in Azure


mbarton:
NPDA currently follows our reference architecture for configuration in Azure. There is a storage account with a file share containing the `.env` file. We then point the container to it using the native [Azure Container Apps support](https://learn.microsoft.com/en-us/azure/container-apps/storage-mounts?tabs=smb&pivots=azure-cli#azure-files).

Azure Container Apps does not yet support accessing the storage account using an Azure managed identity: https://github.com/microsoft/azure-container-apps/issues/850. As we were just using it for config, we decided access using static keys is OK as the data contained there is API keys for third party validation services etc. In particular, the session secret key is not in the `.env` file but injected as an Azure Container Apps secret.

Azure has flagged our usage of storage accounts as critical with two warnings:

- Storage accounts should restrict network access using virtual network rules
- Storage accounts should prevent shared key access

We can fix the first warning standalone by integrating a virtual network into our Azure Container Apps environment but in my experience that tends to cause knock on complexity accessing other resources like the database. I also don't think that warning makes sense if you have shared key access disabled, as that's ultimately the same access control mechanism that the Azure portal is using.

In particular, just restricting access to the storage account to a particular network breaks access from the Azure control panel unless you happen to be within that network. That is unacceptable for how we are structuring our applications - it is vitally important that the members of our team can access and debug our file shares regardless of where they are.

Looking in to this has also made me realise we probably need to disable shared key access entirely - as we now store uploaded CSV data in a separate share to the config. This is ultimately the same data we save in the database and as such needs to be subject to the same access control (Active Directory/managed identity authentication).

So potential way forward:

- Download the `.env` file at container startup, similarly to how we do in [continuous integration](https://github.com/rcpch/national-paediatric-diabetes-audit/blob/d30f8f4c4b950db489b7b0d82d6effafca492284/s/ci#L6)
  - This should use the managed identity of the container, in the same way as we do for database access
- Store the CSV files as blobs in the database (Django [BinaryField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#binaryfield))
 - We previously rejected this as it would be surprising to developers not familiar with the project
 - Django strongly advises against doing this but their arguments apply to commonly accessed static files whereas we are storing the CSV for the record. It is unlikely to be accessed again, certainly not regularly and only by authorised users.
 - Prevent shared key access and suppress the warning about virtual network rules

[...]

PR #546, opened 2025-01-31
Store CSV uploads in the database

mbarton:
First part of the fix for #498. Store CSV files in the database rather than on disk. This is because we can't restrict access to the Azure File Share using time limited credentials, which is our red line for access to patient data.

Rather than also store the data quality spreadsheet in the database I've changed it to generate on demand.

I thought about using `defer` to avoid loading the CSV file into memory when you load the `Submission` but so far I can't see it obviously being a performance issue. Will fix if it comes up in #538


project/npda/migrations/0021_submission_csv_file_name_alter_submission_csv_file.py
```diff
@@ -0,0 +1,30 @@
...
+        migrations.AddField(
+            model_name="submission",
+            name="csv_file_name",
+            field=models.CharField(
+                help_text="Name of the uploaded CSV file",
+                null=True,
+                verbose_name="CSV file name",
+            ),
+        ),
+        migrations.AlterField(
```
mbarton, line 22: Suprisingly this is a backwards compatible change, existing rows end up with the old CSV filename as the CSV data. We should clear out staging just to avoid confusion.

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 39 seconds ago) Please check your changes!

PR #552, opened 2025-02-02
486-transfer-patient

eatyourpeas:
Adds transfer fields to the patient form

As yet no work flow for hiding or flagging transferred patients to the users


eatyourpeas:
Screen Recording 2025-02-02 at 21 47 35

eatyourpeas:
Just a few questions about this.

**Questions:**
1. if the transfer is saved, should the patient appear still in the list for that submission (I think yes). If so, should it be signposted in someway to the users that a transfer has been made?
2. if the transfer has been saved, should we then disable further editing of the record?

eatyourpeas:
image

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 40 seconds ago) Please check your changes!

Issue #539, opened 2025-01-31
Create prod instance


mbarton:
Once #498 is done

Also todo:

- [ ] enable emails on error logging

mbarton:
~~Instance stood up and running but failing the healthcheck due a strange error about “request.user” not existing on WSGIRequest. Not sure why this is happening on live but not staging~~ I was a noodle and had mistyped the allowed hosts setting

PR #554, opened 2025-02-05
Fix running Djangos native createsuperuser

mbarton:
Part of #539. I found I couldn't create myself as the first superuser because we have overridden `create_superuser` but not allowed some of the arguments that the native Django call passes. Rather than pass `**kwargs` I have whackamoled each argument as it returned an error. This might break again in future versions of Django, but we are unlikely to use the native command apart from when we create brand new environments.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 38 seconds ago) Please check your changes!

PR #566, opened 2025-02-08
Rearrange async code to fix httpx pool timeouts

### LINKED ISSUE #560: Pool timeout uber crash when uploading real data
@AmaniKrayemRCPCH and I tried a real live CSV file in the live environment. Unfortunately the upload rather destroyed the server instance, flooding my email with timeout errors from the HTTPX pool.

My suspicion is that the all the patients had the same postcode in previous load tests. So the response from our postcodes.io instance was pretty zippy, presumably because of the db page cache. Real data is of course spread across multiple postcodes and I think those requests being slower then leads to the pool timeouts (default 5 seconds).

I have the file so can test more but have successfully uploaded our test `dummy_sheet.csv`.

Some actions:

- Check if my theory is correct: measure latency for a "warm" postcode lookup vs a cold one
- We can probably pooch the current implementation to "work" by reducing the number of rows processed in parallel
  - At the moment it still does all the rows at once and hopes the async pool can take up the slack
  - I think batching is preferable than fiddling with the pool timeout but if we're thinking of batching...
- Think about implementing Celery etc now (#515) so I get single job failures in my inbox not thousands of errors!

mbarton:
Fixes #560 

Before we tried to run validation tasks for ALL the patients in parallel, using a task group for each patient.

That led to HTTPX pool timeouts with real data - the default is 5 seconds.

This rearranges the async handling to do bounded parallelism at the patient level - ie only handle max 5 patients in flight at once.

The old task group was actually redundant as we now parallelise lookups in external_patient_validators and external_visit_validators using asyncio.gather


project/npda/general_functions/csv/csv_upload.py
```diff
@@ -155,34 +130,11 @@ def retain_errors_and_invalid_field_data(form):
...
-                tasks.append(task)
-
-        return [task.result() for task in tasks]
-
     def record_errors_from_form(errors_to_return, row_index, form):
         for field, errors in form.errors.as_data().items():
             for error in errors:
                 errors_to_return[row_index][field].extend(error.messages)
 
-    def do_not_save_patient_if_no_unique_identifier(patient_form):
```
mbarton, line 173: Moved to `Patient.save`

project/npda/models/patient.py
```diff
@@ -148,8 +147,7 @@ class Meta:
             CAN_OPT_OUT_CHILD_FROM_INCLUSION_IN_AUDIT,
         ]
 
-    def clean(self):
-        super().clean()
+    def save(self, **kwargs):
         if not self.nhs_number and not self.unique_reference_number:
```
mbarton, line 153: Don't save at all if id fields are missing rather than just flagging it as an error

project/npda/tests/test_csv_upload.py
```diff
@@ -366,11 +423,6 @@ def test_invalid_nhs_number(test_user, single_row_valid_df):
     errors = csv_upload_sync(test_user, single_row_valid_df)
     assert "nhs_number" in errors[0]
 
-    # Catastrophic - Patient not save
```
mbarton, line 399: It's fine to save an invalid NHS number, it will be flagged in the data quality report and UI

project/npda/general_functions/csv/csv_upload.py
mbarton: Moved back as this caused unrelated test failures

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 46 seconds ago) Please check your changes!

PR #569, opened 2025-02-09
Support DD/MM/YY (short year) for CSV uploads

### LINKED ISSUE #568: Support diagnosis date with short year format
NB: I only managed to find this bug when testing #566 - without that you get timeouts which hide this error

The real test CSV file from the previous platform has a different format for diagnosis date: DD/MM/YY rather than DD/MM/YYYY. As we specify `error=coerce`, this means none of the patients save as they have no diagnosis date (it's set to `NaT`).

We do specifically mandate DD/MM/YYYY in our guidance: https://www.rcpch.ac.uk/sites/default/files/2024-02/npda_dataset_2021_guidance_updated_feb_2024.pdf. But it's not realistic to expect everyone to change and we need to upload files from the old platform to validate the behaviour of the new one.

I think we need to fix multiple things here:

- Support DD/MM/YY
- Errors parsing datetime columns should surface immediately (ie `errors=raise` not `errors=coerce`).

mbarton:
Fixes #568. DD/MM/YY seen in the wild in an uploaded file from last year

### DIFF
--- a/project/npda/general_functions/csv/csv_parse.py
+++ b/project/npda/general_functions/csv/csv_parse.py
@@ -120,7 +120,7 @@ def csv_parse(csv_file, is_jersey=False):
 
     for column in ALL_DATES:
         if column in df.columns:
-            df[column] = pd.to_datetime(df[column], format="%d/%m/%Y", errors="coerce")
+            df[column] = pd.to_datetime(df[column], format="mixed", dayfirst=True, errors="coerce")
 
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 6 minutes and 37 seconds ago) Please check your changes!

PR #588, opened 2025-02-10
validation-patient-nhs-number

### LINKED ISSUE #585: validation for duplicate NHS Numbers in questionnaire
It is still possible to save patients with duplicate NHS numbers in the questionnaire.

eatyourpeas:
Fixes #585 

Add validation in the clean method to the patient form
- if nhs_number is present, check if does not already exists
- if urn is present, check if does not already exist
already exists

In both of these circumstances it raises a validation error. 

**this now breaks most of the tests because it causes an async error**

Other things:
1. fix for the headers in the patient table. Reorganises the filtering so that those patients that have not completed a full year or died now come at the bottom and have a dark blue header
2. fix the hovertext console non-critical error relating to pyplot
3. reorganise the deprivation colours so that pink and red are deprived, blue and green are not

### DIFF
--- a/project/npda/forms/patient_form.py
+++ b/project/npda/forms/patient_form.py
@@ -198,6 +198,31 @@ class PatientForm(forms.ModelForm):
                 ),
             )
 
+        if nhs_number:
+            if (
+                Patient.objects.filter(nhs_number=nhs_number)
+                .exclude(id=self.instance.id)
+                .exists()
+            ):
+                self.add_error(
+                    "nhs_number",
+                    ValidationError(
+                        "A patient with this NHS number already exists in the database."
+                    ),
+                )
+        if unique_reference_number:
+            if (
+                Patient.objects.filter(unique_reference_number=unique_reference_number)
+                .exclude(id=self.instance.id)
+                .exists()
+            ):
+                self.add_error(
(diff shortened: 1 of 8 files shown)


eatyourpeas:
@mbarton not sure if this is going to work if we validate in this way? In this PR I have added ValidationErrors for existing NHS Numbers or URN. This works fine in the questionnaire but now breaks the CSV upload with an async error the fails in this new validation step. Is it that it thinks each row is a new patient, even through it is only the visit that is new?

mbarton:
Ah i see. The error is just that you can’t use sync database calls in an async function. Django mandates this to avoid inadvertent blocking.

We could maybe do some sync_to_async wrapping as we have elsewhere. But I’m wondering if it’s better just to get on and implement Celery. With multiple workers we would  not need async at all.

eatyourpeas:
I agree - do you want to pair on it on Thursday? I added `aexists` in the form as an experiment and it worked except then it broke further down as `nhs_number` was now missing from `cleaned_data`.

mbarton:
This is blocked on #613

mbarton:
I think this was covered in https://github.com/rcpch/national-paediatric-diabetes-audit/pull/732

PR #621, opened 2025-02-17
Handle view_preference more securely

### LINKED ISSUE #612: Newly registered users can see all data
1) Start a fresh instance
2) Add a patient to RCPCH PZ999 as the default superuser
3) Register a new user to a different PDU
4) Login as that user
4) Add a patient to that PDU

To begin with you won't see any patients as you don't yet have a submission. But after the submission is created, you will see all the patients.

This happens because new users are set with a `view_preference` of `0` - organisation level. But that is no longer respected in the code calculating the patient queryset for the list view:

https://github.com/rcpch/national-paediatric-diabetes-audit/blob/a043ecd11105c15c23d001c841206d07fee7ca6a/project/npda/views/patient.py#L114

Security critical code like this **cannot** be an opt-in check. The default path should be the most restrictive. 

As important as fixing this bug is understanding why our current permissions and visibility checking tests did not catch this. It may be the case they're missing entirely: https://github.com/rcpch/national-paediatric-diabetes-audit/pull/191 only added them for the user view as we didn't have any patient seeding functions yet.

mbarton:
Fixes #612 

- Disallow setting `view_preference` other than 1 or 2
  - Fixes security issue where you could construct a malicious request to set view_preference to something larger than 2 and see everything
- Always check `is_rcpch_audit_team_member` alongside `view_preference=2` for defence in depth
- Remove view preference HTMX callback in users view - it's now handled in the `view_preference` route

### DIFF
--- a/project/npda/general_functions/view_preference.py
+++ b/project/npda/general_functions/view_preference.py
@@ -1,5 +1,5 @@
 import logging
-from django.core.exceptions import PermissionDenied
+from django.core.exceptions import PermissionDenied, BadRequest
 from django.apps import apps
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 6 minutes and 47 seconds ago) Please check your changes!

PR #650, opened 2025-02-21
Second-stick-of-celery

### LINKED ISSUE #515: CSV upload progress bar
At the moment when you upload a CSV file it just sits there and thinks. We know it’s doing something because we can see the logs but that’s frustrating behaviour to normal users, especially as we expect our biggest uploads to take 30-60 seconds to process.

We should introduce a progress bar. The simplest version of this would be a fake one that moves up by itself using JavaScript but I think we can do better.

We could use [HTMX Server Sent Events](https://htmx.org/extensions/sse/) support with Django [StreamingHttpResponse](https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.StreamingHttpResponse) to return the true progress as well as some text describing what is happening.

Note that when running under WSGI Django will consume the entire async streaming response before returning it. This wouldn’t work for us. We could move to ASGI but I’m a bit scared since none of our other code is asynchronous. 

As a middle ground we could have an [async queue](https://docs.python.org/3/library/asyncio-queue.html) between the async upload code and the response. This would mean more CSV uploads than WSGI workers would hang the server until they complete but that’s probably fine for us in practice.

If it’s not or we find the solution above too hacky in practice we can add Celery (and Redis) containers to process the upload in jobs. We could then show progress through simple polling of a progress field on the Submission model.

eatyourpeas:
Closes #515 

### Overview
Implements celery / redis / celery beat
Adds 3 new containers so requires update to env
Will need similar changes within azure before merging

removes the async functions in the csv_upload function but leaves in the async API calls to other services


eatyourpeas:
Just to say these tests are all passing, so this must be to do with the configuration, maybe the environment variables not existing in the action.

There are 4 tests that I have skipped which cover missing header name errors of various types. This is because the `csv_upload` function is now broken into 2 and testing database outcomes here nolonger quite makes sense.

eatyourpeas:
ok, the remaining tests are rewritten and tests now passing.

eatyourpeas:
ScreenRecording2025-02-23at22 23 39-ezgif com-video-to-gif-converter

project/npda/general_functions/csv/csv_upload_celery.py
```diff
@@ -0,0 +1,227 @@
...
+    # Process each patient and their visits
+    tasks = []
+    for patient_index, patient_group in visits_by_patient:
+        patient_row = patient_group.iloc[0]
+        patient_dict = row_to_dict(patient_row, Patient, csv_headings=CSV_HEADINGS)
+
+        print(f"Submission ID: {submission.id} before saving patient")
+
+        patients_submission_task = save_patient_and_visits_to_submission.s(
+            patient_row_json=patient_row.to_json(date_format="iso"),
```
mbarton, line 205: I didn't realise DataFrames supported pickling: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_pickle.html.

Could we avoid recleaning the dates in the Celery task? (`convert_invalid_dates_to_none`, `convert_iso_dates`). If Celery doesn't support sending binary data to a task we can base64 encode the pickling.

project/npda/tasks.py
```diff
@@ -0,0 +1,352 @@
...
+    patient_row = pd.Series(convert_iso_dates(deserialized_patient_row))
+    patient_group = pd.DataFrame(patient_group_dict)
+    patient_form = PatientForm(data=patient_dict)
+    pdu = PaediatricDiabetesUnit.objects.get(pz_code=pz_code)
+    submission = Submission.objects.get(id=submission_id)
+
+    # This protects the whole block of code from failing if there is an error and the database cannot be rolled back
+    # This is important because we are creating multiple objects in the database
+    # Hopefully it will never be needed, but it's good to have it just in case
+    with transaction.atomic():
```
mbarton, line 229: Wouldn't this remove the patient if one of their visits has a catastrophic error saving?

project/npda/tasks.py
```diff
@@ -0,0 +1,352 @@
...
+                if not visit_form.is_valid():
+                    retain_errors_and_invalid_field_data(visit_form)
+                    record_errors_from_form(
+                        errors_to_return, visit_row["row_index"], visit_form
+                    )
+                    # if the errors are critical (eg type mismatch) we should cast these values to None
+                    # so that we can continue with the row
+                    for field, errors in visit_form.errors.as_data().items():
+                        for error in errors:
+                            print(f"Critical Error {error.code}: {error.messages}")
```
mbarton, line 304: Should be in the line below?

project/npda/tests/test_csv_upload.py
```diff
@@ -217,55 +349,131 @@ def test_multiple_patients(
...
+def test_missing_date_of_birth(single_row_valid_df, mocked_submission, test_user):
+    # Delete all patients to ensure we're starting from a clean slate
+    Patient.objects.all().delete()
+
+    single_row_valid_df.loc[0, "Date of Birth"] = None
+    assert (
+        Patient.objects.count() == 0
+    ), "There should be no patients in the database before the test"
+
+    submission = Submission.objects.create(
```
mbarton, line 356: Do we need both this and mocked_submission?

project/npda/general_functions/csv/progress_recorder.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def set_progress(self, current, total, patient_id):
+        progress_data = {
+            "patient_id": patient_id,
+            "current": current,
+            "total": total,
+            "percent": (current / total) * 100,
+            "errors": 0,
+        }
+        self.redis_client.set(self.task_id, json.dumps(progress_data))
```
mbarton, line 27: Do we need some namespacing around the task ID?

project/npda/tasks.py
```diff
@@ -0,0 +1,352 @@
...
+            if "model" in entry and apps.get_model("npda", entry["model"]) == model:
+                model_field_name = entry["model_field"]
+                model_field_definition = model._meta.get_field(model_field_name)
+
+                csv_value = row[entry["heading"]]
+                model_field_value = csv_value_to_model_value(
+                    model_field_definition, csv_value
+                )
+
+                if model_field_name in [
```
mbarton, line 81: My gut feeling is that we would do this in `read_csv` instead? Ideally whatever dataframe comes out of that can be processed without additional cleaning

mbarton (commented):
Thanks @eatyourpeas for plugging away at this.

I'm wondering about breaking this PR down a bit to ease roll out? Probably one PR to introduce Redis and the new containers and then another one to change the upload code.

I'm also wondering if it's over engineering at this point to try and support both uploading processing it all in-route and with a flag to use Celery? We will almost certainly need to fix bugs with this that only crop up testing once merged and it would be nice to have that option in our pocket.

I guess the blocker is the existing code is async? From memory we tried sticking `async_to_sync` at a high level and it didn't work.

mbarton:
Also in the spirit of deploying these changes as risk free as possible, I wonder if we track progress on the Submission model itself? Something like a total patients from CSV and total visits from CSV which the progress checking endpoint then compares by querying the number of patients, visits and `__all__` errors under the submission itself.

My gut feeling is that until we have other use cases that need it Redis should just be there as an implementation detail of Celery and not be required for our features. In theory we could swap to something else without changing other code, especially if Celery finally get round to allowing Postgres as the queue (https://github.com/celery/celery/issues/5149)

mbarton:
Celery uploads are now supported and we will roll out to all users.

Merging the additional refactoring work done in this PR (Celery task per patient) is tracked under https://github.com/rcpch/national-paediatric-diabetes-audit/issues/974

Issue #687, opened 2025-03-04
[LIVE - SUNDERLAND ROYAL] Cannot add additional patients using questionnaire


mbarton:
@orangepeel73 was testing the questionnaire and found he couldn't add additional patients as the "Add Patient" button was missing on the patient list.

![Image](https://github.com/user-attachments/assets/02ed1dba-3efc-46d0-869b-e277a3b96631)

FWIW - I can see the Add Patient button so I suspected a permissions issue. But to see no UI at all `view_preference` needs to be set to national. Neil's view preference is set to `1` - trust. So I am thoroughly confused about what's happened here.

https://github.com/rcpch/national-paediatric-diabetes-audit/blob/b44e50936b45c3d53c9fac4f8e9d5d847264f036/project/npda/templates/partials/page_elements/patient_table_actions.html#L37

![Image](https://github.com/user-attachments/assets/7eae1090-4c0e-4163-90f0-af76d93c50c9)

eatyourpeas:
I have just seeded a fresh instance, created a new user who is editor access and a clinician, and selected questionnaire and created a patient. I get this:
<img width="1510" alt="Image" src="https://github.com/user-attachments/assets/b230ea1d-1604-469e-b722-e13182bfd90b" />
The add patient button looks to be there. I have checked his permissions in the admin and they all look to be correct.

mbarton:
This seems fixed now

PR #692, opened 2025-03-04
Speculative fix for missing add patient button

mbarton:
Speculative fix for #687. I can't reproduce the issue but the only way it could be happening is if the `view_preference` is an unexpected value. So change the comparison to hopefully handle that, by hiding the UI if the view preference is national rather than only showing it if it's trust (in reality PDU).

### DIFF
--- a/project/npda/templates/partials/page_elements/patient_table_actions.html
+++ b/project/npda/templates/partials/page_elements/patient_table_actions.html
@@ -34,7 +34,7 @@
             </button>
         </label>
     </label>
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 52 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 1 second ago) Please check your changes!

PR #707, opened 2025-03-05
height-weight-mandatory

### LINKED ISSUE #705: [FORM VALIDATION] height and weight both required as well as date
In our analyses, for BMI, we use the most recent visit where both height and weight were completed. If one was completed without the other, we wouldn’t use that measurement.
I guess there will be many instances where only one is done, so we may not want to return an error(?), but users need to know that if they haven’t got a concurrent height and weight, then the BMI health check will be coded as incomplete. (@AmaniKrayemRCPCH )

Currently in the questionnaire we only require the date and one of either height or weight. Should I enforce that both are required therefore?

eatyourpeas:
Fixes #705

### Overview
Feed back that height and weight together mandatory with observation if one of them supplied

### DIFF
--- a/project/npda/forms/visit_form.py
+++ b/project/npda/forms/visit_form.py
@@ -775,13 +775,29 @@ class VisitForm(forms.ModelForm):
             measure_must_have_date_and_value(
                 height_weight_observation_date,
                 "height_weight_observation_date",
-                [{"weight": height}],
+                [{"weight": weight}],
+            )
+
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 55 seconds ago) Please check your changes!

PR #709, opened 2025-03-05
ketone-meter-sick-day-training

eatyourpeas:
### Overview
Sick day training and ketone meter training are not the same thing, though often occur together. The first requires an assertion (Y/N/Unknown), the second requires a date.

This PR separates the need for the two measures to exist together as previously. It removes the validation requiring the two to be completed together and any other associated tests.

It also reverses the previous PR around height/weight validation as per [this reply](https://github.com/rcpch/national-paediatric-diabetes-audit/issues/705). Height/Weight are now not mutually exclusive and the tests enforcing this have been removed.

A tooltip is added to remind users that BMI cannot be scored without both on the same date. Note this generates a new migration


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 8 minutes and 1 second ago) Please check your changes!

PR #724, opened 2025-03-06
truncated-tooltips

eatyourpeas:
### Overview

This PR makes safe all tool tips so that any html codes for quotes etc are rendered as intended. There was a tendancy in the lists for tooltips to be truncated at the top of the list and by making them show below this problem is overcome.

In the process I realized as a superuser it is not clear which organisation's patients you are looking at so I have added the PDU lead organisation name and PZ code to all headers of all tables and all forms


eatyourpeas:
image
image

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 56 seconds ago) Please check your changes!

Issue #488, opened 2025-01-16
Invalid date of birth error causes "this field is required"


mbarton:
Edit `dummy_sheet.csv` and change the date of birth for the first patient to `26/10/` (invalid formatting).

The error in the data quality report says Date of Birth: This field is required.

I think it should say invalid date.

mbarton:
There's a variation on this issue for visit date - it ends up as None and no error is highlighted

PR #752, opened 2025-03-12
Null out invalid fields before saving in CSV uploads

### LINKED ISSUE #719: [LIVE - UCL] Test handling of invalid data for all CSV columns
Write tests to handle the columns having incorrect values. Pull out the code from #650 that catches save errors and nulls out the fields to ensure the model still saves.

The aim of this ticket is that we are confident in April people will not experience a catastrophic error uploading and they should get as many patients and visits as they have rows in the CSV, assuming they haven't messed up required fields.

mbarton:
Fixes #719 

Port across the behaviour from #650 where we null out invalid fields to avoid exceptions at the database layer.

Follow up tasks:

- #488 to handle invalid dates. Errors must be created in csv_parse as we end up with `NaT` for both invalid dates and blank cells

TODO:
  - [x] diagnosis date test
  - [x] ~~should we have a TransferForm?~~ reworked so `PatientForm` is handling transfer fields as in the questionnaire


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 53 seconds ago) Please check your changes!

PR #755, opened 2025-03-13
Git info loads above navbar if branch loaded is not live

### LINKED ISSUE #745: Identify if you're on local dev or staging instances
We had a test user mistakenly add patient data to staging. I also switch a lot between instances and I'm scared I'll do the same.

We should prominently display which instance you're on, unless it's live. In E12 we also display git hash to superusers which is nice.

dc2007git:
## Overview

Utilises code from E12 to print the git branch, specific commit, and logged in user above the navbar in NPDA. This info is only shown if user is a superuser.

Also add an instance label which is shown for everyone to help identify whether you're on local dev, staging or live (where it's not shown at all).

Screenshot 2025-03-22 114301

## Related Issues
Closes #745


mbarton:
Thanks @dc2007git, this is great! Unfortunately now we no longer have separate branches for `staging` and `live`, we'll need something else to indicate which instance you're on.

I guess it could be a config value? In previous projects I've had a config value called `label` which is then prominently displayed next to the logo

mbarton:
@dc2007git thanks again for this! I've made a couple of tweaks:

- Show the dev info on any branch but only for superusers
  - There are some unlikely situations where we'll need to deploy a branch that isn't `live` so it's good to have the dev info remind us of this in case we forget
- Add an `INSTANCE_LABEL` config that's shown as a label next to the the home button on the nav
  - I'll configure it to show `staging` but hidden in live

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @dc2007git and merged by @mbarton 6 minutes and 46 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (created by @dc2007git and merged by @mbarton 7 minutes and 53 seconds ago) Please check your changes!

Issue #741, opened 2025-03-12
[LIVE - BLACKPOOL] Missing columns when uploading CSV


mbarton:
![Image](https://github.com/user-attachments/assets/649cf533-0e9d-4cf3-9c68-5a77f7c99010)

I've asked them to send us their CSV file with just the headers to debug, as the platform won't store the CSV file if it hits this error (as a security measure against storing data we don't need).

mbarton:
We've also had the same error reported by Manchester University NHS Foundation Trust. They've sent me the CSV file and some of their column names don't match our template:

- `NHSNumber` (expected `NHS Number`)
- `DOB` (expected `Date of Birth`)
- `Effective Death Date` (expected `Death Date`)
- `Visit Date` (expected `Visit/Appointment Date`)

~~So the issue here is probably that the error message is showing correct columns when it should show incorrect columns.~~ the error message is correct just impossible to read. It's actually split into "missing columns" and "unexpected columns", so affected columns appear twice in the message.

@AmaniKrayemRCPCH @richardshepherd45 do you know if the column names they're using are from an older template we need to support?

richardshepherd45:
Hi Michael,

Thanks for the update.

I’ve discussed with Amani and we believe that the new platform should not support the old CSV template. The new one has been available for quite a while, and they’ll be able to download the new one directly from the platform.

Hopefully, for those using the old CSV this will mean just a change in the names of the columns. The number of columns is the same so it shouldn’t be too difficult for them to switch over.

I can go back to the Manchester University NHS FT and let them know that they are using an old template and that this will have caused an issue with uploading the data.

Kind regards

Richard

mbarton:
OK @richardshepherd45 sounds good. I will still update the platform to give a more readable error when this happens - at the moment it's very garbled and hard to understand

mbarton:
Reopening as the CSV file for Manchester comes from a separate system and it would be a development change on their end to modify it. It's easier for us to just accept the alternate column names

PR #757, opened 2025-03-13
Fix crash when NHS number column missing

mbarton:
In #741 the user was using an old template where the `NHS Number` column was `NHSNumber` instead. This led to a crash, now fixed.

I wrote a few more unit tests while I was trying to understand the problem so have committed them here too.


mbarton:

Issue #761, opened 2025-03-14
'View user logs' tooltip just shows my name for all users


AmaniKrayemRCPCH:
![Image](https://github.com/user-attachments/assets/101bc624-3346-4805-b31e-b7502d2972f3)

Regardless of who I hover over, it says 'View user logs for Ms Amani Krayem'

anchit-chandran:
main character energy @AmaniKrayemRCPCH

anchit-chandran:
Fixed in #762

Issue #756, opened 2025-03-13
[LIVE - LANCASHIRE, ROYAL ALEXANDRA] Patient characteristics totals too high


mbarton:
PZ183 ROYAL PRESTON HOSPITAL has 261 patients but the Patient Characteristics graph says XXX / 1516
PZ135 ROYAL ALEXANDRA CHILDREN’S HOSPITAL has 270 but the graph says XXX / 1687

anchit-chandran:
Believe this happens due to the Q object generated in `kpis.py::177`:

```
self.patients = Patient.objects.filter(
            visit_falls_within_audit_period_Q_object(
                prepend_query_path="visit", audit_start_date=self.audit_start_date
            ),
            paediatric_diabetes_units__paediatric_diabetes_unit__pz_code__in=pz_codes,
            patientsubmission__submission__submission_active=True,
            patientsubmission__submission__submission_date__range=(
                self.audit_start_date,
                self.audit_end_date,
            ),
        )
```

A chatgpt suggested fix that seems to work is simply add `.distinct` on the end

# SQL

**RAW SQL generated by the whole query**:

```
>>> print(self.patients.query)

[...]

PR #762, opened 2025-03-14
[KPIS] fix calculations

anchit-chandran:

This PR does 2 important, small & significant changes, which apply across every kpi test file (leading to a large diff):

1. Revert the additional filter inside the base patient query that asserts a patient to be included in kpi calculations must have at least 1 Visit which includes at least valid measurement date within the audit period. The KPI query definitions just look a `Visit.visit_date` as the base query; and each individual calculation will individually check measure dates where applicable.
2. Update tests to all add Patients to a Submission first. Tests were previously passing silently, even though kpi calculations required filtering on valid Submissions, because they directly set the `.patients` attribute (as they were made a couple months before the Submission model was in place) and I missed the change. Validated by first using the correct calculateKPIS' public method to set patients (used `set_patients_for_calculation(pz_codes=[])` as this includes the valid submission query), leading to expected fails for every test. Then, for each test, create a Submission and add patients, fixing them again. It's the exact same change across the 40+ test files contributing to the large diff.

**Additionally**:
- removes all references to the old `calculate_kpis_for_single_patient` method from code and docs
- fixes tiny UI bug #761 
- autoflake,isort, linting etc 

### Related Issues

#756 
#761


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @anchit-chandran and merged by @mbarton 6 minutes and 56 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (created by @anchit-chandran and merged by @mbarton 8 minutes and 3 seconds ago) Please check your changes!

PR #766, opened 2025-03-17
Support old NPDA template columns

### LINKED ISSUE #741: [LIVE - BLACKPOOL] Missing columns when uploading CSV

mbarton:
Fixes #741 by supporting columns from an older version of the NPDA template.

We decided to do this as some units would have to go back to the developers of their reporting script to change how they are generated. It's easier for us to support the old format as the columns are the same, just with different wordings, typos or sneaky non breaking spaces


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 3 seconds ago) Please check your changes!

Issue #797, opened 2025-03-21
Percentages don't add up for closed loop, pump and CGM in dashboard


mbarton:
It's possible that the percentages shouldn't add up but it's still confusing.

Feedback from Justin is that this entire section perhaps could just be numbers rather than percentages. They may also want to remove the HCL graph entirely.

@mbarton to gather more feedback before we can pick up this issue

mbarton:
Royal Preston have also reported that the overall counts don't match their expectations. I'm having a look now

mbarton:
The percentages don't add up because it's still using the KPI calculations. Taking the Closed Loop number as an example, that's saying that 93% of patients with insulin pump as their *treatment* are using a closed loop system. Whereas the Pump number is the percentage who have insulin pump as their treatment overall.

It's still confusing to me that the former number should ever be higher than the latter, as it appears to be in the Royal Preston data.

mbarton:
I think the problem with the counts comes because the KPI calculation code blindly uses the most recent visit regardless of whether that visit has any data about the treatment.

I can reproduce the problem by adding two patients manually using the questionnaire and then adding a visit each:

1) On pump, closed loop
2) On pump, not closed loop

At this point, the numbers correctly show Closed Loop: 1, Pump 2. But if you then add another visit to one of them that doesn't record treatment, the pump number will snap back to 1.

mbarton:
I've double checked the Royal Preston closed loop numbers with a pivot table on the original data and they are correct

eatyourpeas:
so am coming round to thinking about using the graph like this:

![Image](https://github.com/user-attachments/assets/e0ae0af8-0aaf-4057-bab2-00f96dc9d6bd)

This would be the total (pump starts, new diagnoses, dka admissions, transitions) cumulatively month on month for the whole submission. The card would show the total numbers to date in this submission, clicking on the card would cause a histogram to appear in the graph card with maybe numbers of cases by quarter, with a line graph to show the weekly or monthly trend. What do you think?

eatyourpeas:
In follow up this is the idea:

<img width="714" alt="Image" src="https://github.com/user-attachments/assets/94fbce6a-5718-4e8a-94c9-c89650f5e337" />

There are only a few metrics where cumulative data are interesting and how they fluctuate quarter by quarter - these are my suggested choices.

The small cards would continue to show the totals as absolute values, the graph would show the rate of change over the submission quarter to quarter.

[...]

PR #809, opened 2025-03-22
Fix incorrect pump counts on dashboard (KPI 15)

mbarton:
Part of #797 

Royal Preston noticed that their closed loop count was higher than their pump count. This shouldn't happen, as the patients on a closed loop are a subset of the patients on pumps.

This was caused because the calculation for KPI 15 (patients on pumps) was looking for the most recent visit, regardless of whether a treatment type was registered for it. That led to undercounting. The code for KPI 24 (closed loop) did not have this problem as it already had a predicate on the treatment type.

This PR adds `treatment__isnull=False` to the subquery that finds the most recent visit.

### DIFF
--- a/project/npda/kpi_class/kpis.py
+++ b/project/npda/kpi_class/kpis.py
@@ -1310,7 +1310,8 @@ class CalculateKPIS:
         # Define the subquery to find the latest visit
         latest_visit_subquery = (
             Visit.objects.filter(
-                patient=OuterRef("pk"),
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 54 seconds ago) Please check your changes!

PR #814, opened 2025-03-22
Fix KPI 24 calculations for patients who no longer use HCL in their most recent visit

mbarton:
Fixes a bug I saw in the Royal Preston data where the denominator for Closed Loop on the dashboard was not the same as the numerator for Pump.

The bug was caused by a patient who was on an HCL but had gone back to injections in their most recent visit.

The fix is to use the exact same code we already use for KPI 15, which correctly applies the filter on treatment type only when calculating `passed` patients rather than in the subquery to find the most recent visit with a treatment record.

### DIFF
--- a/project/npda/kpi_class/kpis.py
+++ b/project/npda/kpi_class/kpis.py
@@ -1756,12 +1756,9 @@ class CalculateKPIS:
 
         # Eligible kpi24 patients are those who are either on an insulin pump or insulin pump therapy
         eligible_kpi_24_latest_visit_subquery = (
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 47 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 6 minutes and 55 seconds ago) Please check your changes!

PR #819, opened 2025-03-23
organisation-employer-fixes

### LINKED ISSUE #817: Administrators should not be able to see superuser flags
Although site administrators cannot set is staff, superuser and other flags, they can see the labels.
Image

If logged in as an administrator you should only see the fields you can edit.

### LINKED ISSUE #818: Administrators can't edit users with dual PDU membership if they don't have that membership themselves
Not sure if this is totally the problem but this is what I have found.

1. login as superuser and create a user with Administrator role in a PDU
2. create another role in the same PDU as a Viewer and an Editor
3. logout, then confirm the accounts of the administrator in turn of each of the new users and logout
4. Login as the Administrator and try and edit the different users in turn

What should happen
1. the adminstrator should be able to edit each of the different users

What does happen
1. the administrator cannot edit the Editor account and gets a 403

eatyourpeas:
### Overview
There were a few issues in the NPDA user form that are addressed by this PR:

## User form

1. labels relating to superuser / admin status are now hidden from view if users are not rcpch users or superusers
2. check boxes in the form have been changed for daisyui toggles
3. users that have edit access but are only affiliated with one organisation but are not superusers/RCPCH members nolonger can see the add employer button (since they cant add organisations to users that they are not themselves members of).
4. removes Children and Families from the role dropdown

## User table

1. Users in the user table who are associated with more than one organisation now have a hospital icon next to them.
2. if the administrator does not have access to the users second organisation, they cannot edit them so now this is signposted in a tool tip and the click through is disabled - the NPDA will be able to add them as they have extra privilege
3. Table UI signposting - the tooltips on the user id now explicitly say if you can edit the user and render the user id as a button to make this clearer


eatyourpeas:
image
and
image
and
image
and
image

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 6 minutes and 48 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 9 seconds ago) Please check your changes!

PR #863, opened 2025-03-31
Fix duplicate org employer re-running user import

mbarton:
Fixes a bug where you'd get duplicate `OrganisationEmployer` entries for users if you ran the same import multiple times.

The first PDU seen for a given user becomes the primary. This is not ideal as we don't really know whether the user considers that their primary place of work (where they want to go when they initially log in). But I think on balance it's better to preserve the implicit invariant that each user has one primary org employer to avoid confusion for future developers!

Also fixes the logging message at the end to correctly log the number of affected users.

### DIFF
--- a/project/npda/management/commands/import_users.py
+++ b/project/npda/management/commands/import_users.py
@@ -50,7 +50,10 @@ class Command(BaseCommand):
 
         with open(file_path, "r") as file:
             reader = csv.DictReader(file)
+            
             index = 0
+            unique_emails = set()
+
             for row in reader:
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 6 minutes and 53 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 2 seconds ago) Please check your changes!

PR #872, opened 2025-04-01
Allow password reset for inactive users

mbarton:
as that's how we imported them from the CSV

PR #873, opened 2025-04-01
Test PR to debug CI failures

mbarton:
Do I see the same test failures as in https://github.com/rcpch/national-paediatric-diabetes-audit/pull/872?


mbarton:
We do see the same failures, closing this

PR #864, opened 2025-03-31
Create test users for non-admin roles

mbarton:
Automatically creates additional non-admin users when running locally.

Refactors import_users command to use the same user creation logic as creating the dev users.

PR #874, opened 2025-04-01
Revert "Create test users for non-admin roles"

mbarton:
Reverts rcpch/national-paediatric-diabetes-audit#864


mbarton:
I raised this speculatively to see if #864 caused the issue with the tests failing but it didn't.

PR #903, opened 2025-04-02
view-npda-user-can-view

### LINKED ISSUE #898: Readers cannot view patient form
Readers get a 403 when clicking through to the patient form. I think they need to be able to see the data just not edit it?

### LINKED ISSUE #899: Changing role in user list does not update group/permissions
1. Create a Reader
2. Change the Reader to be a Coordinator

Expected: they are able to edit data
Actual: they can't

If you go in and fix the group in the Django admin it works. So looks like just a bug in our user list. Tagging as LIVE - UCL as it affected @FrancescaAnnan who was initially registered as an editor but needed to be a coordinator.

### LINKED ISSUE #900: Changing role in user list does not update group/permissions
1. Create a Reader
2. Change the Reader to be a Coordinator

Expected: they are able to edit data
Actual: they can't

If you go in and fix the group in the Django admin it works. So looks like just a bug in our user list. Tagging as LIVE - UCL as it affected @FrancescaAnnan who was initially registered as an editor but needed to be a coordinator.

eatyourpeas:
### Overview

The permission on the NPDAUser UpdateView was set to change_npdauser, but that prevented those with view only access from gaining access. So this PR drops the access to the NPDA update form to view access but then disables the fields in the form for them. There is a belt and braces guard in the update form as well to raise a 403 if the form is submitted by someone without the `change_npdauser` permission.

In effect therefore there should be no change in the security.


eatyourpeas:
this will also close #899 and #900

Latest commit adds a `form_valid` method to the NPDAUser UpdateView which clears the groups from the user and adds the group associated with the new role.

I am pretty happy that all this now works but this is about authentication so checking @mbarton if could run locally and confirm all good to go.

## summary of changes
1. updates group of user if role is updated in a new override of `form_valid` in the `NPDAUserUpdateView`
2. downgrades NPDAUserUpdateView permission to view_npdauser access from edit_npdauser access. A check instead for the `edit_npdauser` permission now happens in the new `form_valid` method. The existing permission in the NPDAUserCreateView are left untouched.
3. All fields are disabled for view_npdauser permissions in the create/edit user template
4. Those with view only access can now navigate to the form from the NPDAUser list without causing a 403 error. (the word 'edit' is removed from the tooltip for them to save questions later).
5. incidentally the NPDAUser list now has new fontawesome icons

mbarton (approved):
This looks fine. I would have liked a test for the 403 we now raise manually, will add one whilst addressing #906

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (created by @eatyourpeas and merged by @mbarton 8 minutes and 1 second ago) Please check your changes!

PR #910, opened 2025-04-03
can_upload_csv_permission

eatyourpeas:
### Overview

This PR fixes the current vulnerability for readers to upload csvs. 

It introduces two new permissions to the NPDAUser model (generating a new migration) and checks against this for the logged in user and raises a PermissionDenied if the permission is not in the user group. A test for this has been added also.

In addition links and buttons are disabled in the templates with messaging where appropriate to signpost to users that they do not have permissions to upload csvs

##NOTE that after merging, while the group seeding and `group_for_role` functions have been updated, the new permissions will need manually adding to the administrator, editor and superuser groups in the admin


project/constants/user.py
```diff
@@ -117,6 +117,16 @@
     "Can consent to participating in NPDA.",
 )
 
+CAN_SUBMIT_CSV = (
+    "can_submit_csv",
+    "Can submit a CSV file to NPDA.",
+)
+
+CAN_DOWNLOAD_CSV = (
```
mbarton, line 125: Do we check this anywhere?

eatyourpeas (commented):
So this PR now ready to merge.
Items included in this PR:
1. includes implementation of the `can_download_csv` and `can_submit_csv` permissions in the model, view and group seeding.
2. checks for the same in all the templates
3. add checking for `delete_submission` permission which had no implementation previously. This now is done in the view and the template.
4. tidy the submission list

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 8 minutes and 1 second ago) Please check your changes!

Issue #909, opened 2025-04-03
Any logged in user can view user logs for a known user ID


mbarton:
1. Create two coordinator users in different PDUs
2. Visit the user logs page in one PDU
3. Manually edit the URL to the primary key for the user in the other PDU

Expected: 403
Actual: you see the user logs

The fix should be straight forward, stick `CheckPDUInstanceMixin` on `NPDAUserLogsListView`.

For what it's worth I don't think Readers and Editors should be able to see user logs apart from for themselves?

PR #934, opened 2025-04-07
[User Logs] Update perms

anchit-chandran:
- editors/readers cant see any other users' logs
- coordinators can see logs of users in own pdu
- rcpchauditteam can see everyone

@mbarton @eatyourpeas could you confirm this behaviour is fine and implementation as expected for the VisitActivity user log view permissions?

I ended up not putting inside `CheckPDUMIxin` as that relies on introspecting the CBV's model being used - so would have to add another check inside the mixin `if model == VisitActivity` and move the logic i've put directly inside the `NPDAUserLogsListView.has_permission` method into the mixin. But this felt not right as it's ONLY used in this view

#909


project/npda/views/npda_users.py
```diff
@@ -431,9 +434,68 @@ class NPDAUserDeleteView(
...
+            return False
+
+        try:
+            npda_user = NPDAUser.objects.get(pk=npdauser_id)
+        except NPDAUser.DoesNotExist:
+            return False
+
+        # If user is a Reader or Editor, they can only see their own logs
+        logged_in_user: NPDAUser = self.request.user
+        is_reader_or_editor = logged_in_user.role in [
```
mbarton, line 462: For other user related actions we have tended to check permissions directly rather than role. Eg `npda.change_npdauser`.

I'm not sure it's justified to add another permission for this but maybe we check the existing permission instead for consistency?

I'm torn between the options. What do you think?

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @anchit-chandran 7 minutes and 57 seconds ago) Please check your changes!

Issue #753, opened 2025-03-13
Connect to Redis using Entra ID


mbarton:
Convert this command line for use in the app:

```
export REDISCLI_AUTH=$(curl -H "x-identity-header: ${IDENTITY_HEADER}" "${IDENTITY_ENDPOINT}?resource=https://redis.azure.com&api-version=2019-08-01/" | jq -r .access_token)
redis-cli --user <MANAGED_IDENTITY_ID> -h <REDIS-NAME>.uksouth.redis.azure.net -p 10000 --tls
```

Guidance here: https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/managed-redis/managed-redis-entra-for-authentication

We will also need to periodically get a new token and send `AUTH` commands down the connection. I suspect this might need us to create the connection ourselves and pass it to Celery.

mbarton:
Going to park this for now as Entra ID authentication against the instances using the CLI has stopped working AGAIN

Clearly this functionality is not stable on the Azure end so we'll continue to use a static key for now but store it as an Azure Container Apps secret as we do with the Django secret (another static credential we can't avoid).

We can come back to this ticket once Microsoft make Managed Redis GA and see if it's reliable then.

PR #943, opened 2025-04-08
Redis/Celery settings for Azure

eatyourpeas:
### Overview
This adds function for the django app and the celery instance to authenticate (and communicate) with azure redis

changes:
1. Adds a `CACHES` redis backend to `settings.py`
2. pulls `REDIS_HOSTNAME`, `REDIS_PORT` from `.env` locally if running locally. If `AZURE_CONTAINER_ENVIRONMENT` is present in the environment (as is default in Azure), instead the variables are set from the azure config. The password is pulled from the azure credential library
3. A similar set up in `celery.py` but includes a workaround as redis instances in azure often work like clusters (for internal sharding and so on) and setting a global key prefix allows celery redis broker operations to be handled together within azure.
4. adds django-redis to `requirements.txt` as a dependency to handle serialization/deserialization from celery to redis and back

For this to work
1. make sure that your local .env has `REDIS_HOSTNAME`, `REDIS_PASSWORD` and `REDIS_PORT` are present. Note the this has changed from `REDIS_HOST`
2. On deployment this will deploy presumably to staging and live? Possibly needs a conditional to make sure this currently only happens on staging?

### DIFF
--- a/project/celery.py
+++ b/project/celery.py
@@ -6,6 +6,8 @@ from celery import Task
 from django.conf import settings
 import django
 
+from azure.identity import DefaultAzureCredential
+
 # Logging setup
 logger = logging.getLogger(__name__)
 
@@ -27,6 +29,22 @@ app.conf.update(
         "global_keyprefix": "{rest}:"
     }
 )
+
+def get_redis_token():
+    credential = DefaultAzureCredential()
+    redis_resource_id = "https://redis.azure.com"
+    token_object = credential.get_token(redis_resource_id)
+    return token_object.token
+
+broker_host = os.environ.get("REDIS_HOST")
+broker_port = os.environ.get("REDIS_PORT", 6379)
+broker_password = get_redis_token()
+broker_username = os.environ.get("MANAGED_IDENTITY_OBJECT_ID") # Set this as an environment variable
+
+broker_url = f"redis://{broker_username}:{broker_password}@{broker_host}:{broker_port}/0"
+
+CELERY_BROKER_URL = broker_url
+CELERY_RESULT_BACKEND = broker_url # If using Redis as result backend
 
 app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
 
--- a/project/celery.py
+++ b/project/celery.py
@@ -22,30 +22,26 @@ app.config_from_object("django.conf:settings", namespace="CELERY")
 broker_url = os.environ.get('CELERY_BROKER_URL')
 result_backend = os.environ.get('CELERY_RESULT_BACKEND')
 
-
-REDIS_HOST = os.environ.get('REDIS_HOST')  # Your Redis hostname (e.g., <your-cache>.redis.cache.windows.net)
-REDIS_PORT = os.environ.get('REDIS_PORT', '6379')
-REDIS_PASSWORD = os.environ.get('REDIS_PASSWORD')  # Your Redis password
 # Local development Redis configuration
-# Default Redis configuration for local development
+REDIS_HOSTNAME = os.environ.get('REDIS_HOSTNAME')  
+REDIS_PORT = os.environ.get('REDIS_PORT', '6379')
+REDIS_PASSWORD = os.environ.get('REDIS_PASSWORD')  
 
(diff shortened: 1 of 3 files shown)


project/celery.py
```diff
@@ -16,17 +18,47 @@
...
+        result_backend = f"redis://:{REDIS_PASSWORD}@{REDIS_HOSTNAME}:{REDIS_PORT}/1"
+        app.conf.update(
+            broker_transport_options={
+                "global_keyprefix": "{queue}:"
+            },
+            result_backend_transport_options={
+                "global_keyprefix": "{rest}:"
+            }
+        )
+    except Exception as e:
```
mbarton, line 53: I would not try and handle this, it's a fatal exception if this doesn't work and we don't want to bring that container instance into service at all

project/settings.py
```diff
@@ -344,17 +345,34 @@
...
+# Conditional configuration for Azure environment
+if "AZURE_CONTAINER_ENVIRONMENT" in os.environ:
+    REDIS_PASSWORD = None # set to None as we will get it from Azure AD
 
-CELERY_BROKER_URL = f"{redis_protocol}://{redis_auth}{REDIS_HOSTNAME}:{REDIS_PORT}/{REDIS_DATABASE_NUMBER}{redis_params}"
+    credential = DefaultAzureCredential()
 
-CELERY_RESULT_BACKEND = CELERY_BROKER_URL
\ No newline at end of file
+    try:
```
mbarton, line 371: This appears to duplicate the main celery config, worth refactoring out into a function?

project/celery.py
```diff
@@ -16,17 +18,47 @@
...
+# Conditional configuration for Azure environment
+if "AZURE_CONTAINER_ENVIRONMENT" in os.environ:
+    REDIS_PASSWORD = None # Set to None as we will use Azure AD token
+
+    credential = DefaultAzureCredential()
+
+    try:
+        token = credential.get_token("https://redis.azure.com/.default")
+        REDIS_PASSWORD = token.secret
+        broker_url = f"redis://:{REDIS_PASSWORD}@{REDIS_HOSTNAME}:{REDIS_PORT}/0"
```
mbarton, line 43: We must use ssl here `rediss://` and `?ssl_cert_reqs=required` otherwise we would be potentially passing patient information across the internet in cleartext

project/celery.py
```diff
@@ -16,17 +18,47 @@
...
+REDIS_HOSTNAME = os.environ.get('REDIS_HOSTNAME')  
+REDIS_PORT = os.environ.get('REDIS_PORT', '6379')
+REDIS_PASSWORD = os.environ.get('REDIS_PASSWORD')  
+
+broker_url = f"redis://{':' + REDIS_PASSWORD + '@' if REDIS_PASSWORD else ''}{REDIS_HOSTNAME}:{REDIS_PORT}/0"
+result_backend = f"redis://{':' + REDIS_PASSWORD + '@' if REDIS_PASSWORD else ''}{REDIS_HOSTNAME}:{REDIS_PORT}/1"
+
+
+# Conditional configuration for Azure environment
+if "AZURE_CONTAINER_ENVIRONMENT" in os.environ:
```
mbarton, line 35: If we make this our own setting in `.env` we can conditionally enable in staging

requirements/requirements.txt
```diff
@@ -41,6 +42,7 @@ gunicorn>=22.0.0
 celery==5.4.0
 flower==2.0.1
 redis==5.2.1
+django-redis==5.4.0
```
mbarton, line 45: Is this required to do tasks with parameters? If so, let's update the test task to use it.

mbarton (changes requested):
Thank you @eatyourpeas - this seems like a good step in the right direction.

Happy to have a call to discuss further or shall I pick this up?

requirements/requirements.txt
mbarton: We think tasks with parameters will work without this - I will try without and put back in if we need it

mbarton:
Parking for now, as per https://github.com/rcpch/national-paediatric-diabetes-audit/issues/753#issuecomment-2794102439

PR #954, opened 2025-04-11
Celery flavoured refactor of the CSV upload code

### LINKED ISSUE #336: Delete old submission (including patients) after CSV upload successful
Now we can have multiple patients with the same NHS number (#324) we can be a bit safer in how we handle CSV uploads.

If we move the code to delete the old submission to after we have fully validated the CSV and written all the data, an unhandled error during the process would lead to the new submission being inactive but with the old one preserved. With the behaviour as it stands you'll end up with only the potentially half uploaded new submission.

I wonder if for full safety we want to delete the old submission after a given period of time rather than straight away? This is potentially easier said than done given we don't have a periodic scheduled task system set up yet.

We certainly want to retain the last but one CSV upload.

mbarton:
Split creating the submission, processing the file and tidying up old submissions into separate methods so they can be called in different places in the future (http route and Celery tasks).

Add the VisitActivity entry as soon as we've created the submission so we still have a record of the upload even if it crashes.

Delete patients from old submissions after we have successfully processed the new one to make it easier to revert back in case of a crash processing the new file. (Fixes #336).

Import models directly rather than using `apps.get_model`.

### DIFF
--- a/project/npda/general_functions/csv/csv_upload.py
+++ b/project/npda/general_functions/csv/csv_upload.py
@@ -32,24 +32,69 @@ from project.npda.forms.visit_form import VisitForm
 from project.npda.forms.external_patient_validators import validate_patient_async
 from project.npda.forms.external_visit_validators import validate_visit_async
 from project.npda.general_functions.csv.csv_clean import csv_clean
+from project.npda.models import (
+    Patient, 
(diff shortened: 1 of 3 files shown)


mbarton:

PR #961, opened 2025-04-15
Upload CSV into existing questionnaire submissions

mbarton:
Follow up to #959 as some questionnaire units have already added some of their patients manually to the new system.

NB: the Jersey upload does not currently work. I will raise a follow up PR for that to avoid blocking uploading patients to the other PDUs.

PR #959, opened 2025-04-14
Import data from CSV for multiple PDUs

mbarton:
Part of #919 

Adds `--import-as-questionnaire-entries` to the `upload_csv` command. This allows us to import data exported from the old platform as if they had been entered using the questionnaire on the new platform, saving a whole bunch of typing for users.

It creates a new submission for each PDU based on the `PDU Number` column in the CSV. If any PDU already has a submission, no data is uploaded.

This PR also fixes the `upload_csv` command, which I broke in #954.

PR #962, opened 2025-04-15
Fix import of Jersey questionnaire patients

mbarton:
Follow on to #959 and #961 

Refactors `csv_parse` to not require `is_jersey` as a parameter. This means we can use it as part of a bulk upload, where we use the `PDU Number` column to determine PDU. It's a bit silly since only Jersey uses CSV data with `Unique Reference Number` but keeps things consistent.

Also skip calculating IMD for Jersey as it's not supported yet.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes ago) Please check your changes!

PR #967, opened 2025-04-23
Add timeout to celery uploads

mbarton:
Make sure a crash that totally nerfs a celery upload doesn't lead to the user forever seeing the loading screen by adding a 10 minute timeout, after which it flashes a message and logs an error so we get an admin email.

The user still won't see their data as the new submission is probably not marked as active but that's maybe a good thing since we won't know if it's partial data or not.

### DIFF
--- a/project/npda/views/submissions.py
+++ b/project/npda/views/submissions.py
@@ -1,7 +1,8 @@
 # Python imports
-from datetime import date
+from datetime import date, datetime, timezone
 import json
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 52 seconds ago) Please check your changes!

PR #973, opened 2025-04-25
Support terminated postcodes

### LINKED ISSUE #716: Terminated postcodes
The test upload from UCL had one postcode that failed validation. It is a valid postcode just not in use any more. We need to fall back and call the postcodes.io terminated endpoint if we can't find the postcode in the main API endpoint initially. You still get location information back.

mbarton:
Fixes #716 

This PR also removes a duplicate call to the postcodes.io API endpoint. We were calling it first to validate the postcode, then again to get the location. Now we return the location from the first call.

Finally it fixes a bug where we would try to lookup the IMD for all postcodes.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 43 seconds ago) Please check your changes!

PR #976, opened 2025-04-26
Issue972-search-users

### LINKED ISSUE #972: Search users
Can we add a search bar to the manager users view?

Often users will email us with a query, but not specify their unit, so currently we have to either:
* figure out their unit and then filter 
* look for them in the unfiltered list

Would ideally need to search by email address and name.

eatyourpeas:
### Overview
Implements a search bar for NPDA users
We have django-filter in the dependencies already so it is used for this

### Code changes
- creates a new filterset 
- changes the NPDAUserList to a FilterList
- Adds a search input to the template similar to the patient list

### DIFF
--- a/project/npda/filtersets/__init__.py
+++ b/project/npda/filtersets/__init__.py
@@ -1,0 +1,1 @@
+from .npdauser_filterset import *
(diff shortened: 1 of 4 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 39 seconds ago) Please check your changes!

PR #987, opened 2025-05-05
save-invalid-postcodes-in-questionnaire

### LINKED ISSUE #985: Override postcode validation failure for questionnaire submission
We've had a patient added via questionnaire with a postcode that is new enough not to appear in the postcodes.io database we use for validation. We've advised the user to add the patient with a dummy postcode for now that we will then update later.

This will get painful as we have more examples of this (another one appeared in E12 today).

We think it makes sense to allow users to override the postcode validation failure for questionnaire submission. Probably a button next to the validation error that says "Accept anyway". This could then set a hidden form field that skips the validation on the server side.

We should still trigger the admin email (via logger.error) to send us the invalid postcode to check for sure.

The patient naturally will not have an IMD score as we can't determine their location. Some options:

- Accept it and fill them in at analysis time manually (we did this with E12 cohort 5)
- Fill them in using a periodic job (so they are set once the postcode appears in the dataset after upgrading our postcodes.io instance)

eatyourpeas:
### Overview

Fixes #985

## The problem
The ONS postcode list is a little behind new builds, so new postcodes flag as invalid when they are actually valid. This is preventing patients from being saved in the questionnaire.

## The solution
To have a two-step validation process on form submission.
### First submission
1. user enters 'invalid' postcode and submits form
2. form fails cleaning and returns error in the postcode field. Also renames submit button to "Save Changes with Invalid Postcode Anyway" together with an alert box explaining why the postcode is invalid and how it will not block saving if the user is sure.
### Second submission
1. User submits the form and the postcode is saved.


eatyourpeas:
image

mbarton (approved):

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 48 seconds ago) Please check your changes!

PR #992, opened 2025-05-07
Hospital discharge date not mandatory

### LINKED ISSUE #991: Admissions - Discharge date not necessary
We'll accept a hospital admission w/o a discharge date, as they may still be in hospital at the time of data entry.

We can keep this as an error, in case it's accidentally left blank, but can we allow users to save the questionnaire even if the discharge date is missing

mbarton:
Fix #991

### DIFF
--- a/project/npda/forms/visit_form.py
+++ b/project/npda/forms/visit_form.py
@@ -1037,7 +1037,6 @@ class VisitForm(forms.ModelForm):
         if any(
             [
                 hospital_admission_date,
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 50 seconds ago) Please check your changes!

PR #1006, opened 2025-05-20
Support ISO-8859-1 CSV files

### LINKED ISSUE #999: CSV decode error
> Invalid CSV format: 'utf-8' codec can't decode byte 0xa0 in position 2328: invalid start byte

I suspect this happens as Excel's default behaviour is not to save in UTF-8, although there is an option to do so. I'm not sure exactly what encoding the file causing the error is using as we don't save it if it fails.

We could try editing the dummy sheet to add some Latin-1 characters and see if we then get the same error? I have tested full Unicode characters and they just come out as `???`.

mbarton:
Fixes #999 

Unit reported a `UnicodeDecodeError` when uploading their file. Whilst we wouldn't expect any of our columns to contain non-ASCII characters there was a sneaky non-breaking space in an empty Death Date cell.

On a UK English machine Excel will save a CSV with non-ascii characters as ISO-8859-1 (Latin-1). It is possible to select the separate UTF-8 option but I think it's better for us to just fall back if UTF-8 decoding fails.

### DIFF
--- a/project/npda/general_functions/csv/csv_parse.py
+++ b/project/npda/general_functions/csv/csv_parse.py
@@ -66,7 +66,14 @@ def csv_parse(csv_file):
     lowercase_headings_list = [heading.lower() for heading in HEADINGS_LIST]
 
     # Read the first row of the csv file
-    df = pd.read_csv(csv_file)
+    try:
+        df = pd.read_csv(csv_file, encoding="utf-8")
+    except UnicodeDecodeError:
+        # This is the default you get from Excel when saving on a UK English machine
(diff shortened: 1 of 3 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 46 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 57 seconds ago) Please check your changes!

PR #1011, opened 2025-05-26
API

### LINKED ISSUE #923: API
I'm not sure we have an issue to track that we still want to stand up an API so EPRs like Twinkle can submit data directly

eatyourpeas:
### Overview

This implements an API. 

#### Highlights:
1. add django-oauth-toolkit

To do this usually there is a process of swapping for migrations within the package and the migrations in the application. To avoid this we have a custom `PDUAccessTokenProfile` model which wraps the library's AccessToken model. This allows us to store the user's PDU and admin status. There is documentation and examples on how to create an access token and use it with curl or in postman. 

2. DELETE methods are not supported. That will continue within the platform or in the admin

3. The token permission scopes are`patient:read` and `patient:write` as well as `admin:cross-pdu`. These are applied in an authentication class and in the viewsets.

4. urls are `api/v1/patients/` for GET / POST / PATCH and PUT and `api/v1/patients/<id>/visits/`
for the same.

The id relates to the NHS number URN depending on the PDU in the token scope. Note that only the patient instance associated with the current submission in the current audit year is returned.

#### Other things

PDU scoping only allows for a single PDU. I don't see why  an API user should have access to more than one PDU for retrieving and posting data.

There is provision for token renewal but this is not tested.

#### Documentation

There is documentation that I will continue to add to

#### Tests

Obvs there are lots of these


eatyourpeas:
Viewable at: `https://npda.localhost/api/v1/schema/docs/#/`

image

eatyourpeas:
The reason the tests stopped passing here was an identified incompatibility in django-factoryboy with pytest 8.4 which is recently released. See issue here: https://github.com/pytest-dev/pytest-factoryboy/issues/232

eatyourpeas:
@mbarton I think I forgot that in here there is also some changes around searches against `audit_period` and `audit_year` highlighted by these branch conflicts. I am not sure without review whether this fixes what you wanted to achieve with the underlying issue of audit years. We could run through together if you wanted and maybe cherry pick these commits?

eatyourpeas:
failing test here are coming in from live - they are not from this branch. Best fix them in live and then merge into this.

PR #1017, opened 2025-06-07
String-int-visit-error

### LINKED ISSUE #1015: height and weight fields should strip out any units included
Item posted from Frimley:

Image

This happens because cm / kg have been submitted alongside the float in the csv field.

Fix involves fixing the csv parse function to strip non-empty height and weight fields of additional strings.

### LINKED ISSUE #1016: Invisible trailing spaces in 'empty' cells in the csv causing errors
Highlighted by users in Salisbury.
An invisible space in otherwise empty rows in the `Only complete if   DKA selected in previous question: During this DKA admission did the patient   receive any of the following therapies?` column leads the other DKA fields to raise an error.

The fix for this is to strip all empty fields of extraneous spaces.

eatyourpeas:
## Overview
This PR fixes a csv parsing vulnerability highlighted by two different PDUs. It also addresses a new dependency issue in pytest-django

Two users (Salisbury and Frimley Park) have identified parsing errors in the csv. These are strictly user errors, rather than bugs in the platform but highlight a need for NPDA to be more flexible with csv parsing.

Items reported were:
1. invisible white space in free text fields was generating errors when dependent fields were empty (eg DKA admission fields require a date and a reason for admission to pass validation)
2. height and weight fields expect floats but the user was submitting the measurement with the units so this parsed as a string

The solution to these was to apply more cleaning functions to the `clean_csv` file.

### Code changes
1. height and weight fields return floats or None (if empty). If the fields are supplied as a string, all characters except the numbers and the decimal place are removed and returned as a float. If the whole string has no numbers, that is still returned to be cleaned and rejected downstream in form validation.
2. fields with only whitespace in all other rows have that stripped

Tests have been added for both of these

### Pytest Versions
This is documented in the API branch where this was first noticed as a problem.
Recently `pytest-factoryboy` became incompatible with `pytest-django` as the `pytest` version inside it (8.4) broke the way `pytest-factoryboy` handled fixtures. This is documented in this [github issue](https://github.com/pytest-dev/pytest-factoryboy/issues/232). The fix is hopefully fairly temporary - to pin `pytest` to 8.3.0. There is a PR already open to fix the underlying issue.
In the same breath I have pinned django to latest stable (5.2) and bumped `pytest-django` to latest version (4.11.1). Mostly selfishly as this is what I have done in the API PR and it will prevent merge conflicts later on.

### DIFF
--- a/project/npda/general_functions/csv/csv_clean.py
+++ b/project/npda/general_functions/csv/csv_clean.py
@@ -42,10 +42,8 @@ def clean_csv_measurement(value):
         # Remove any non-numeric characters except for decimal points
         # This will handle cases like "5.9cm", "70kg", etc.
         # It will also remove any leading or trailing whitespace
-        print(f"Original value: '{value}', {type(value)} {value is None}")
         value = ''.join(char for char in value if char.isdigit() or char == '.')
     try:
-        print(f"Converting value to float: {value}")
         return float(value)
     except ValueError:
         return None
--- a/project/npda/general_functions/csv/csv_clean.py
+++ b/project/npda/general_functions/csv/csv_clean.py
@@ -23,13 +23,41 @@ def clean_csv_sex(value):
             # Return Not Known. The patient won't save in the database without a numeric value.
             return 0
 
+def clean_csv_measurement(value):
+    """
+    Convert measurement values to float, handling empty strings and non-numeric values.
+    Remove extraneous non-numeric characters if necessary.
+    Returns None for empty strings or non-numeric values.
+    If the value is a valid number, it will be converted to float.
+    If the value is an empty string, it will return None.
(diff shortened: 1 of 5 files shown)


anchit-chandran:
image
additionally adds pz code next to org employer labels for #1018

anchit-chandran (approved):

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (created by @eatyourpeas and merged by @anchit-chandran 7 minutes and 36 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (created by @eatyourpeas and merged by @anchit-chandran 8 minutes and 40 seconds ago) Please check your changes!

PR #1031, opened 2025-06-19
achondroplasia-measurement-override

### LINKED ISSUE #1030: Add override function for height/weight DQ checks to allow valid outlier submissions
A user flagged a valid entry for a child with skeletal dysplasia being blocked due to the current height SD limit (> -8 SD). We’d like to implement an override function allowing users to save entries like this, with a warning prompt.

- All BMI fields must still be submitted to count the care process as complete.
 
- These flagged records can then be excluded from centile/SDS analysis but retained for care process reporting.

Thanks

Richard

eatyourpeas:
### Overview
Some patients with diabetes have disorders which mean we cannot calculate their measurement centiles using UK-WHO. However, providing no measurements would mean the PDU would fail their care process for this measure, so the audit team have asked for an override button in the visit form to allow 'invalid' measurements to be saved anyway.

### Code changes
Introduction of a new `overrride_height_weight` flag in the visit form which conditionally shows an alert in the template and persists the 'invalid' measurement on second submit.

incidentally adds RCPCH name to the PDU list in the select

### Related Issues
closes #1030

### DIFF
--- a/project/npda/forms/visit_form.py
+++ b/project/npda/forms/visit_form.py
@@ -129,6 +129,7 @@ class VisitForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         self.patient = kwargs["initial"].get("patient")
+        self.override_height_weight = kwargs.pop("override_height_weight", False)
         super(VisitForm, self).__init__(*args, **kwargs)
         for field_name, field in self.fields.items():
             model_field = Visit._meta.get_field(field_name)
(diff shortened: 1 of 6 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @eatyourpeas 7 minutes and 30 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 48 seconds ago) Please check your changes!

PR #1036, opened 2025-06-22
Issue1022-eatyourpeas-duplicate-patients

### LINKED ISSUE #1022: It is possible to add the same patient twice to the same PDU
Is this behaviour we should allow?
If one PDU adds a patient who exists in the same submission in another PDU, users are notified in the patient list. The same is true if one PDU adds the same patient twice.

@AmaniKrayemRCPCH should we prevent users from doing this? Note that if a user enters the same patient twice in the same PDU they will be warned that the patient has a duplicate entry in this submission

eatyourpeas:
### Overview
First noticed by KCH (PZ215) where one patient was entered twice in error in the questionnaire and the user could not delete it.
This led to a conversation with the audit team who advised that a patient (identified by NHS number) can be in duplicated in the same submission if the duplicates are in different PDUs, but cannot be duplicates in the same PDU.

### Code changes
In the course of investigation I discovered that the `get_form_kwargs` method in `PatientCreateView` had been duplicated following introduction of the `postcode_override` so only the second one was being called. This meant that PDU and audit year were not making it into the initialized form, and therefore the patient would always be saved. These methods have now been merged. The update class has been amended also.
I also discovered that it was not possible to delete patients who had validation errors so this has also been fixed.
A test has been added.

### Related Issues
closes #1022

### DIFF
--- a/project/npda/forms/patient_form.py
+++ b/project/npda/forms/patient_form.py
@@ -112,6 +112,7 @@ class PatientForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         self.audit_year = kwargs.pop("audit_year", None)
+        self.audit_period = kwargs.pop("audit_period", None)
         self.paediatric_diabetes_unit = kwargs.pop("paediatric_diabetes_unit", None)
         self.override_postcode = kwargs.pop("override_postcode", False)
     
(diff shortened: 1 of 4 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 37 seconds ago) Please check your changes!

PR #1051, opened 2025-06-27
Add audit period to CSV submissions

mbarton:
Fix mistake from #865 where the new `audit_period` reference field was not added to CSV submissions.

This contributed to an issue where NHS numbers were erroneously flagged as duplicates if a submission already existed in a different audit year (#1045). This PR does not fix that issue but will help to stop it happening for future submissions.

PR #1052, opened 2025-06-27
Filter submission by audit year for duplicate detection

mbarton:
When checking patient unique in submission (when submitting via questionnaire).

The async equivalent of this check (used in CSV upload) already referenced audit_year so this makes them consistent.

Submissions created from CSV upload didn't have audit_period set (#1051) but it didn't matter for this line anyway since this only ran during questionnaire submission.

So this really is just ensuring consistency. Once we reliably have `audit_period` set on all Submissions we could switch both of them over.

### DIFF
--- a/project/npda/forms/patient_form.py
+++ b/project/npda/forms/patient_form.py
@@ -422,7 +422,7 @@ class PatientForm(forms.ModelForm):
 
         filter_kwargs = {
             "submission__submission_active": True,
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 9 minutes ago) Please check your changes!

PR #1085, opened 2025-07-05
Issue1075-eatyourpeas-user-logs

eatyourpeas:
### Overview

closes #1075
closes #1073 

Expands the role of signals and VisitActivity.
New features:
1. Logging enabled for `pre_save` and `post_save` on `NPDAUser` and `OrganisationEmployer` models
2. Signals trap changes in either model and log significant changes as a warning, and for important changes, send emails to admin.
3. All changes also stored in VisitActivity model
4. `NPDAUser` also has new `created_by`, `updated_by` to track which users were created/edited and by whom
5. User log table updated with these new fields.
6. Toggles in NPDAUser form fixed as were not properly functioning

New activities logged and stored include:

```
(UPLOADED_CSV, "Uploaded CSV"),
        (TOUCHED_PATIENT_RECORD, "Touched patient record"), #  - This is not implemented yet
        (CREATED_USER_RECORD, "Created a user record"), 
        (CHANGED_USER_RECORD, "User record changed"),
        (CHANGED_USER_ROLE, "User role changed"),
        (CHANGED_ADMIN_FLAG, "Superuser or admin status changed"),
        (DELETED_USER_RECORD, "User record deleted"),
        (ASSIGNED_USER_TO_PDU, "User assigned to PDU"), 
        (REMOVED_USER_FROM_PDU, "User removed from PDU"),
        (USER_PDU_ROLE_CHANGED, "User PDU role changed"),
```
7. Tests also added for the VisitActivity updates

Although the signals firing is not actively tested (as this requires mocking), the endpoint is for the VisitActivity to be updated, and this is tested.

PR #1090, opened 2025-07-07
Log user, audit year and pz code in request logs

### LINKED ISSUE #1071: Log user, PZ code and audit year in request logs
At the moment we can't answer the simple question, what user has looked at what data from our request logs. This is a challenging security position to be in.

I raised #607 for this previously but then rescoped it to just the error emails. I think we need it on the logs too. Unfortunately it's not an easy fix as we're currently using gunicorn request logging:

https://github.com/rcpch/national-paediatric-diabetes-audit/blob/fb0f3741d2fac5517f7282f8a62e1006168a8752/s/start-prod#L12

I'm not sure you can add context to that manually without putting it in req/rep headers which we don't want to do. We could add custom request logging configuration and middleware (like [django-request-logging](https://pypi.org/project/django-request-logging/) but not including the body).

It's possible to change our URLs toinclude pz code and audit year but that is potentially a very big yak to shave (and wouldn't get us user logging anyway).

mbarton:
Fixes #1071 

Replaces the gunicorn request logs with our own middleware, maintaining the same format so we can search logs before and after this PR.

gunicorn always logged the user as empty, it only worked with HTTP basic auth. The middleware knows about the user so can log it as the third log item.

I've tacked on `audit_year=XXXX` and `pz_code=PZXXX` on the end of the line which maintains compatibility with the old format which didn't have it. I felt the `x=y` format helped readability considering it has to be on the other side of the user agent which is inevitably massive. But of course the real solution here is JSON logging.


mbarton:
~~I'm able to reproduce the test issues on `live` - it seems like maybe a test ordering thing leads to the anonymous user being stored in the thread local storage introduced in #1085. It then crashes trying to log their email. I'm still not really sure why the signals are being triggered with a stale user though.~~ ignore this it looks like I just forgot to clean up the thread local after each request!

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 9 minutes and 3 seconds ago) Please check your changes!

PR #1099, opened 2025-07-09
change post-save to pre-save, comment out VisitActivity log

### LINKED ISSUE #1098: Can't delete users
```
IntegrityError at /npda_users/4/delete
insert or update on table "npda_visitactivity" violates foreign key constraint "npda_visitactivity_npdauser_id_4f49e107_fk_npda_npdauser_id"
DETAIL:  Key (npdauser_id)=(4) is not present in table "npda_npdauser".
```

I think this error comes from the signals? An email is triggered that they've been removed from their PDU but they don't actually delete, I guess because the error reverts anything in the transaction associated with the request.

eatyourpeas:
### Overview
🌶️ [Hot FIX]

### PROBLEM

A recently added signal on the `post_delete` method of the `NPDAUser` model predictably threw an error when trying to log the activity in the VisitActivity model - since this model stores the user that has had a change applied, as well as the user that made the change. In the `post_delete` method, the user has already been deleted and therefore cannot be used to update `VisitActivity`

So the fix for the short term has been:
1. change `post_delete` to `pre_delete` to allow logging and email notification to admin to occur.
2. Comment out the saving of the activity to VisitActivity (since any record created there pre-delete would immediately be removed post-delete through referential integrity).

I am detailing this because after I close this PR, I think it would be helpful to have a conversation about what we do with our users that we delete?

I will open a new issue for this but detailing here for continuity.

closes #1098

### DIFF
--- a/project/npda/signals.py
+++ b/project/npda/signals.py
@@ -230,7 +230,7 @@ def log_user_pdu_assignment(sender, instance, created, **kwargs):
         logger.info(f"PDU assignment updated for user {instance.npda_user.email} at PDU {instance.paediatric_diabetes_unit.pz_code}")
 
 
-@receiver(post_delete, sender=OrganisationEmployer)
+@receiver(pre_delete, sender=OrganisationEmployer)
 def log_user_pdu_removal(sender, instance, **kwargs):
     """
     Log when a user is removed from a PDU.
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 9 minutes ago) Please check your changes!

PR #1097, opened 2025-07-08
Temporarily put back the gunicorn logs

mbarton:
Whilst I investigate why #1090 is not working in live

PR #1114, opened 2025-07-11
Revert "Temporarily put back the gunicorn logs"

mbarton:
Reverts rcpch/national-paediatric-diabetes-audit#1097 and loads `.env` before `logging_settings`. It worked without this in dev because we spray all the env vars from `.env` into the container in the Docker compose file. But in staging and live, we just call `load_dotenv`. So it has to come before `logging_settings` for that to be able to read `ENABLE_REQUEST_LOGGING`.

### DIFF
--- a/project/settings.py
+++ b/project/settings.py
@@ -20,14 +20,15 @@ from dotenv import load_dotenv
 #  django imports
 from django.core.management.utils import get_random_secret_key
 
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 9 minutes and 5 seconds ago) Please check your changes!

PR #1119, opened 2025-07-12
npdausers-never-die-they-only-become-less-active

### LINKED ISSUE #1100: Is there life after deletion?
An issue raised by #1099 (in response to issue #1098 )

If we want to retain history of users being deleted and track all activity (as we are doing in VisitActivity), we would do better to set users to `is_active=False`, rather than allow deletion, as we do with E12. 

Currently, any deleted user will be removed from the database. Anywhere a user is used as a foreign key in a model, referential integrity will delete a related record if the user is later deleted, depending on what is set in the `on_delete` attribute in the model's `META` class. At the moment NPDAUser appears as a foreign key in two places:
1. `Submission`: here the `on_delete` attribute is set to `RESTRICT` which means deleting an NPDA user will be prevented and throw an error, if that user has previously been responsible for creating a submission. However, if the submission is deleted, the NPDAUser will be left untouched.
2. `OrganisationEmployer`: this is a middle table between `NPDAUser` and `PaediatricDiabetesUnit`. Here the `on_delete` attribute for the NPDAUser foreign-key is set to `CASCADE`, which will remove employment record of that user but leave the PaediatricDiabetesUnit record untouched. 

So what is our desired behaviour? Instinctively it feels that what we currently have for Submission is not correct. Our options here are:
1. Change the relationship to `SET_NULL`: this allows the NPDAUser to be deleted and the submission will know longer retain a record of who it was created by. The reverse is also true - if the submission is deleted, the NPDAUser remains unscathed. The downside of this method is that we have no audit trail of who did what to whom.
2. Go with option 1 but add extra fields to the `Submission` model (and maybe `VisitActivity` such as `NPDAUser` `email` so that an audit trail can be retained in some form, even after the related user record has been deleted.
3. Disable deletion of users altogether, but instead make them inactive. This means user never die, they just remain forever in the NPDA serving their patients. Accounts can be reactivated in future, and therefore their relationships with the other tables can be persisted. This is most compatible with an extensive audit trail of what users do in the platform.

Now that I read this back over I realize I am starting actually to sound like Claude 4.0, because I am just about to give a recommendation, which you can probably guess. So, I will go full Claude:

## Recommended Approach ##
I recommend **Option 3**
This ensures that:
- ✅ NPDAUsers are gracefully retired
- ✅ `Submission` and `OrganisationEmployer` records are not deleted or left with `None` values in related fields when users are deleted
- ✅ Historical audit trail remains intact
- ✅ Options remain to extend audit trail to include individual updates to other tables such as `Patient` and `Visit`.

*...comes to...*
sorry. sorry.
let me know what you think.

eatyourpeas:
### Overview
This PR is part of the broader issue of audit trails

Currently it is possible to delete users entirely from the platform, which has had knock-on implications for how the `NPDAUser` foreign keys are handled in related tables.

Instead, so that tracking of which user did what when, the first step is to disable the removal of users in favour of making them inactive if they are nolonger part of an audit team.

### Code changes

This has no implications for migrations or the NPDAUser model, since the `is_active` is basic django. This is helpful since purely by setting the flag to `False` means they cannot log in - this function does not need implementing as it is django out of the box.

1. Remove the `npdauser-delete` path from urls, and the associated class in the view, as well as the `confirm_delete` template.
2. Refactors the delete button in the NPDAUser form to reflect 'Activate' or 'Deactivate', depending on the user state. An alert is added to explain why users cannot be removed.
3. This is an important action, so a modal is introduced to check the user is sure this is what they meant to do
4. Refactor the NPDAUser list view to have headers for active and inactive users.

The same permissions apply so no changes have been made here - to activate/deactivate users uses the preexisting `can_delete` permission. Because we are moving what was a delete function into an update one, the `CheckPDUInstanceMixin` has been updated. The rules are:
1. No user can activate or deactivate themselves
2. No user of any role (even superusers) can activate or deactivate a user where they are members of multiple PDUs. This is still possible in the admin.
3. Coordinators are the only role at trust level that can change user status. They cannot do this for a user in a different PDU.
4. Superusers or RCPCH audit team members can change user status across all PDUs, but not where they are members of more than one - this can be done in the admin though.

Tests and documentation have been updated to reflect these changes 

closes #1100


eatyourpeas:
2E2FE663-8F12-4DB7-8A4F-6DA59C5D5D20
661893D3-58B2-4113-B5FB-0A3CCB69E3F4
image

eatyourpeas:
@AmaniKrayemRCPCH Could I check before we merge this change. For those users that are members of 2 PDUs, coordinators are currently not allowed to inactivate users, as inactivating them in one PDU would inactivate them in another.

Currently in fact in the platform no one can inactivate users of multiple PDUs - it can only be done in the django admin. Is that the behaviour you would like? For users of multiple PDUs we could equally have it that only NPDA  team can deactivate them in the platform. Could you let me know what you would prefer?

project/npda/views/npda_users.py
```diff
@@ -449,6 +448,33 @@ def post(self, request: HttpRequest, *args: str, **kwargs) -> HttpResponse:
                     "npda_users",
                 )
                 return redirect(redirect_url)
+
+            elif "deactivate" in request.POST:
+                # Deactivation pathway - toggle the is_active field of the user. A user can only be deactivated if they are not a superuser or RCPCH Audit Team member.
+                # That of course can happen but for now we will only do this in the admin interface.
+                npda_user = NPDAUser.objects.get(pk=self.kwargs["pk"])
+                success_message = f"{npda_user.email} deactivated successfully."
+                if npda_user.is_superuser or npda_user.is_rcpch_audit_team_member:
```
mbarton: I think audit team members should be able to deactivate other audit team members, not sure the superuser flag is relevant here (anyone with that permission can use the admin if this page doesn't do what they need).

project/npda/views/npda_users.py
```diff
@@ -449,6 +448,33 @@ def post(self, request: HttpRequest, *args: str, **kwargs) -> HttpResponse:
...
+                    raise PermissionDenied(
+                        "You cannot deactivate a superuser or RCPCH Audit Team member."
+                    )
+                if npda_user.is_active is False:
+                    success_message = f"{npda_user.email} successfully reactivated."
+                    npda_user.is_active = True
+                else:
+                    # deactivate the user
+                    if npda_user.is_rcpch_staff:
```
mbarton: I think it's fine to allow RCPCH staff to be deactivated?

project/npda/views/npda_users.py
```diff
@@ -467,50 +493,10 @@ def post(self, request: HttpRequest, *args: str, **kwargs) -> HttpResponse:
...
                 else:
                     raise PermissionDenied("You do not have permission to reset two-factor authentication.")
+
             else:
                 return super().post(request, *args, **kwargs)
 
-        
-
-
-class NPDAUserDeleteView(
```
mbarton, line 476: Feels like deactivation should still require a confirmation screen?

project/npda/general_functions/ip_address.py
```diff
@@ -1,4 +1,6 @@
 def get_client_ip(request):
+    if request is None:
+        return "127.0.0.1"  # Default for tests/signals without request
```
mbarton: I think this would confusing in production logs - can we just leave it as `None`?

project/npda/views/npda_users.py
mbarton: Ah I just saw there's logic around this in the mixin too, do we need this here?

project/npda/views/npda_users.py
eatyourpeas: this happens in the template

mbarton:
Thanks @eatyourpeas. @AmaniKrayemRCPCH is away until next week so let’s hold this PR until she can confirm behaviour

AmaniKrayemRCPCH:
Heya. For users in multiple PDUs, if we de-activate their account, it should apply to all PDUs. 
If we want to remove access from just one PDU, we can change their permissions.  

I guess there's a scenario where a user has access to multiple PDUs, and aa coordinator from PDU A wants to remove a user from their unit, but not deactivate their account all together. Is there a way for them to handle this without contacting us to amend permissions?

eatyourpeas (commented):
Just for clarity, these are the rules.
1. Users can nolonger be deleted by anyone
2. Instead users can be deactivated - this retains their data and the relationships their account with other tables allowing an audit trail of their activity to be retained
3. If inactive, they still show in the users list but under a separate heading.
4. Inactive users in the platform can be reactivated
5. Any attempt in the platform to activate or deactivate users results in a warning modal box of the 'are you sure' type. It will also appear in the logs with the name of the user affected, who did it, date/time and ISP
6. Only coordinators can activate/deactivate users in their own PDU.
7. If a user is a member of other PDUs, a coordinator cannot inactivate them. In fact, they can't edit their record at all.
2. RCPCH Audit Team (and superusers) can activate and inactivate all users, irrespective of how many PDUs they are in. 

I have not touched the other workflow around employers and can raise a separate issue for this.

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 49 seconds ago) Please check your changes!

PR #1121, opened 2025-07-12
New navigation (behind feature flag)

### LINKED ISSUE #532: Add breadcrumb navigation
Follow on from #531 - it would be best if we had breadcrumbs leading down from the "Dashboard" through the patient list, visit and down to the visit form.

With this the top level nav items probably become redundant? Clicking the logo as a logged in user gets you to the dashboard. We would then need a call to action on the dashboard to navigate to the patient list, probably contextual as with the top right UI from #531

### LINKED ISSUE #535: Add data entry call to actions to the dashboard
#531 introduces three data upload call to actions directly in the patient table:

- Select CSV upload or manual data entry if no data present in the audit year
- CSV upload and download data quality report
- Add patient

We should have a similar set of call to actions on the dashboard page. It would be more discoverable than learning you have to click "Submissions" in the top nav bar.

We may also want to tweak the first one to take into account the submission method from the last audit year (ie you used this method in 2024).

### LINKED ISSUE #1109: Improve unit selector UX
Reported by @richardshepherd45 in the NPDA catch up 10/07/25.

It's hard to explain to people how to use the unit selector buried in the "Filters" menu.

We should move that functionality to be big obvious tiles on the homepage. This would also help UX as it's the first page you land on when you log in to the platform.

mbarton:
Fixes #1109
Fixes #535 
Fixes #532 

Replaces the top level navigation items and unit selector in the header with a new homepage where you can select audit year and PDU. The header then shows breadcrumb navigation.

I've added this behind the "new_navigation" feature flag. You need to enable the flag, refresh the page and then click the home button.

Screenshot 2025-08-19 160816

Screenshot 2025-08-19 160840

Screenshot 2025-08-19 160905

As there was no way to navigate from a patient to their visits, I've added that as a button (even if the feature flag is not enabled). I've tried to avoid disturbing muscle memory by retaining the "Save Changes" button where it was on the right hand side and moving "Back to list" next to the new button on the left hand side. I'm very open to feedback about how best to do this!

Screenshot 2025-08-19 191338


eatyourpeas:
Pretty. Breadcrumbs would be nice - daisyui have some [nice ones](https://daisyui.com/components/breadcrumbs/?lang=en). The upload button should be either the RCPCH strong blue or the pink probably.
Does this mean that it is possible to have multiple tabs open with different PDU information in it? How does this affect the PDU selected in the session?

mbarton:
Yes that’s another reason to pause this PR. Those links work by adding a “next_page” query parameter to the existing view_preference switcher.

As you say we can only have one active PDU per session at the moment. The current UI means you can only trip up on that if you interact with an old tab.

But this new one would encourage opening a tab per PDU, which wouldn’t work.

For me the fix is easy - add pz_code as a URL parameter and the corresponding permissions check. It’s an easy fix but touches like ALL the endpoints. So if we decide to do it we’d probably want to roll it out one endpoint at a time over a longer period. We should probably also add audit period ID in there too to avoid having to do the same thing later with that.

Our rolling deploy process would mean we’d have to leave the old endpoints in situ and then remove them in an additional PR once no active container apps revision was referencing them. 

Users could still have old pages loaded in old tabs so feels like one big removal PR would make sense as we’d be confident they would have hit the 2 hour timeout by that point.

eatyourpeas:
If it is a big refactor, maybe it is one for copilot... I think it is something worth doing now as it will make life easier in future

mbarton:
> I think it is something worth doing now as it will make life easier in future

OK I'll start shaving the yak and see how bad it gets **EDIT** it's not seeming that bad so far https://github.com/rcpch/national-paediatric-diabetes-audit/pull/1129

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 41 seconds ago) Please check your changes!

PR #1122, opened 2025-07-12
Use get_submission_for_request where possible

### LINKED ISSUE #533: Single method to get the active submission
At the moment we have a few variations on the same pattern of code:

```python
Submission.objects.filter(
    paediatric_diabetes_unit__pz_code=pz_code,
    submission_active=True,
    audit_year=audit_year,
).first()
```

Sometimes we pass the `pz_code` directly from session data or even `request.POST`. Sometimes we already have a `PaediatricDiabetesUnit` model so use that instead of `__pz_code`. Sometimes we forget the `submission_active` bit and order by submission date to get the last. In CSV upload we do it all using asynchronous ORM `a` methods.

I don't know of any bugs caused by this variation but it would be nice to have a single method to call to do this. Semantically it would be best as an instance method on `PaediatricDiabetesUnit` but I'd be fine with a custom manager method on `Submission`:

```python
Submission.objects.active_submission(pz_code="PZ999", audit_year=2024)
```

That would involve rewriting the least amount of code since access via `pz_code` is the most common pattern.

mbarton:
Closes https://github.com/rcpch/national-paediatric-diabetes-audit/issues/533

Also removes a strange call chain where `get_submission_actions` would receive an `audit_year` that came from the `AuditPeriod` helper only to use it to lookup the `AuditPeriod` again.

And remove a redundant submission lookup in `get_children_by_pdu_audit_year`

### DIFF
--- a/project/npda/general_functions/map.py
+++ b/project/npda/general_functions/map.py
@@ -51,25 +51,11 @@ Functions to return scatter plot of children by postcode
 """
 
 
-def get_children_by_pdu_audit_year(
-    audit_year, paediatric_diabetes_unit, paediatric_diabetes_unit_lead_organisation
-):
+def get_children_by_pdu_audit_year(submission, paediatric_diabetes_unit_lead_organisation):
     """
     Returns a list of children by postcode for a given audit year and paediatric diabetes unit
     """
     Patient = apps.get_model("npda", "Patient")
-    Submission = apps.get_model("npda", "Submission")
-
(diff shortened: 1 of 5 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 55 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 3 seconds ago) Please check your changes!

PR #1117, opened 2025-07-11
Add reset 2fa button

mbarton:
Closes #1096 

RCPCH audit team can reset two factor for users. Users could already do this self service but only if they've been able to log in.

Screenshot 2025-07-11 221920

PR #1127, opened 2025-07-14
Fix logic for resetting two factor authentication

mbarton:
Fix logic error in #1117 so that audit team members can do it as well as superusers


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 5 seconds ago) Please check your changes!

PR #1133, opened 2025-07-17
Speed up dashboard load

mbarton:
Loading the GeoDataFrames each time blocked the server, starving out the patient_measurements and all_patient_charts endpoints.

The shapes won't change after load so simply caching them reduces the overall dashboard load time down from ~10 seconds to around 4.

**Before**
Screenshot 2025-07-17 145001

**After**

Screenshot 2025-07-17 145035

We could load the files in the [Django startup hook](https://docs.djangoproject.com/en/5.2/ref/applications/#django.apps.AppConfig.ready) but that would really slow down automatic refresh for local development. So I'm happy to pay the penalty of the first dashboard load in a new container being slow with subsequent ones faster.

### DIFF
--- a/project/npda/general_functions/map.py
+++ b/project/npda/general_functions/map.py
@@ -2,6 +2,7 @@
 import logging
 import json
 import os
+from functools import cache
 
 # django imports
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 7 seconds ago) Please check your changes!

PR #1138, opened 2025-07-19
Issue1134-eatyourpeas-jersey-maps

### LINKED ISSUE #1134: get_map_partial crashes for Jersey
Seen in live telemetry.

```
Error generating map chart

IndexError
invalid GEOS Geometry index: 0
```

We don't have any mapping data for Jersey yet so no point trying to load the map.

### LINKED ISSUE #1135: All patient unit report filters don't work for Jersey
Image

We have no IMD data for Jersey so deprivation should be hidden but you should still be able to click ethnicity and diabetes type

### LINKED ISSUE #1136: Sorting the patient report doesn't work for Jersey
Try to sort any of the tabs in the patient report by unique reference number. It doesn't work

eatyourpeas:
### Overview
This PR fixes open issues relating to Jersey

#### Maps
We do have a map of Jersey, just that we cannot produce a scatter plot of patients against postcodes as we don't have a service that converts these to longitude and latitude. There was logic to skip the mapping of patients if geolocation data not present but this does not seem to work in live. Possibly the location fields in live are not null.
This fix does more extensive cleaning and checking of the patient location data before trying to map it.

#### Patient Report
The median HbA1c calculation currently filters for visits using `patient__nhs_number`. Clearly that will not work for Jersey. In this PR, the `CalculateKPIs` class is now initialize with an optional `is_jersey` flag which is used to choose which fields to filter in the KPI methods that rely on this.

#### [BUG FIX] Jersey CSV
In the process of testing I discovered that you get an error if you upload a Jersey CSV to an English/Welsh PDU, but not the other way round. The checking for this happens in the `POST` method of the `upload_csv` endpoint, before submissions are created and the CSV is cleaned and parsed. I have a applied a fix here.

#### A word about tests
I did create a dummy sheet for Jersey to test the upload errors but  this is difficult as our tests are not end-to-end; we are using  `upload_csv_sync` as a proxy function that creates a submission before simulating upload. The logic in the view to test that the CSV matches the PDU happens before all this though, so it would be a fairly large rewrite to test just this thing.

### Code changes
- Changes `map.py` to add more cleaning steps before accessing the longitude and latitude for patients before plotting
- `Adds a new flag to the `init` method in `CalculateKPIS` class in `kpis.py`. This is an optional `is_jersey` flag.
- Make sure the class is initialized with the `is_jersey` flag in `patient_report` and `patient_measurements`.
- adds jersey dummy csv to fixtures - not used
- add extra logic to `upload_csv` to prevent Jersey users uploading a csv with NHS Number in the header.
- update the documentation on the KPI class
- fixes the filtering by NHS Number / Unique Reference Number in the patient report - updates the `sort_field` to be agnostic of which is used

closes #1134
closes #1135
closes #1136


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @eatyourpeas 7 minutes and 56 seconds ago) Please check your changes!

PR #1140, opened 2025-07-21
Add login_and_otp_required to view_preference route

mbarton:
It was missing it and should only be used by logged in users. Anonymous users couldn't use it anyway as they have no entry in the database but still better to cover it with the same permissions checks as everything else

### DIFF
--- a/project/npda/views/home.py
+++ b/project/npda/views/home.py
@@ -45,6 +45,7 @@ def download_template(request):
     )
 
 
+@login_and_otp_required()
 def view_preference(request):
     """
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 54 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 20 seconds ago) Please check your changes!

PR #1146, opened 2025-07-24
Separate setting for change notifications

mbarton:
Sync behaviour with E12 https://github.com/rcpch/rcpch-audit-engine/pull/1259

Use separate setting for user notification emails rather than relying implicitly on `ADMINS` and `SITE_CONTACT_EMAIL`

PR #1147, opened 2025-07-24
Fix CHANGE_NOTIFICATION_EMAILS in example

mbarton:
Follow up from #1146 - it's an env file so we can't use square brackets!


mbarton:

PR #1160, opened 2025-07-29
Coordinators can only change email or role for users in exactly the same PDUs as themselves

### LINKED ISSUE #1159: Malicious coordinator can access any PDU their users can access
NB: this is an attack I discovered testing our code - there is no evidence this has happened in the live platform.
 
1. Using the default dev setup, grant `incubator+reader@rcpch.ac.uk` access to an additional PDU (other than PZ999)
1. Change the email for that user to one the "malicious actor" controls
1. Open an incognito window and trigger a password reset
1. Change the password and then log in as the reader userThe coordinator can now see the additional PDU, because they changed the readers email to one they control. They can also upgrade their permission to Coordinator, allowing full read/write access to that PDU.

mbarton:
Fixes https://github.com/rcpch/national-paediatric-diabetes-audit/issues/1159

### DIFF
--- a/project/npda/forms/npda_user_form.py
+++ b/project/npda/forms/npda_user_form.py
@@ -75,6 +75,7 @@ class NPDAUserForm(forms.ModelForm):
         # get the request object from the kwargs
         self.request = kwargs.pop("request", None)
         employer_choices = kwargs.pop("employer_choices", [])
+        restricted_fields = kwargs.pop("restricted_fields", [])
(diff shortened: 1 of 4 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 55 seconds ago) Please check your changes!

PR #1171, opened 2025-07-31
Remove national view

### LINKED ISSUE #1083: The national view isn't useful
Raising this more as a discussion item than anything to action immediately but I don't think the national view is useful as it is, at least on most pages.

On the user list it's annoying as an administrator because you have to switch to national view to search for any user. It feels like this could just show you all the users from all the PDUs you have access to (which is all of them for audit team members). Same for the submissions view.

I guess without it we couldn't search for a given NHS number in the patient list without knowing which PDU they are in. But I'm not sure there's a use case for that and you could do it in the Django admin anyway?

mbarton:
Fixes #1083 

The UI shows you the users you can see, although you still can't edit employers if the user has more than one unless you have `can_transfer_npda_lead_centre`.

In the submissions view you see all the submissions and stats if you are an RCPCH audit team member, otherwise just the submissions for the currently selected PDU.

`view_preference` remains as a field on `NPDAUser` should we need it in the future.

Preparatory work for rolling out #1121 - after this PR we won't need to do data URL versions of the user table


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 50 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 59 seconds ago) Please check your changes!

PR #1174, opened 2025-08-01
Increase upload timeout

mbarton:
15 seconds is aggressive enough that it triggers on uploading the dummy sheet and also the last upload from PZ183. Finger in the air bump to 2 minutes.


mbarton:

PR #1185, opened 2025-08-01
Switcher follows audit period and PZ code in URL

mbarton:
Fixes a bug where if you navigated to a different audit period or PDU not via the switcher (i.e. not updating the underlying session) the switcher would show the one still in session.

There is still a variation on this bug where if you change the URL, navigate to a non PDU data page like "home" or "npda_users" then click on a PDU data nav link like "Patient Report", you end up at the PDU in the session not the one you started at. If you click the back button it works as expected since the changed PDU is in the URL. It's only if you navigate via a page that doesn't have a PDU in the URL that the bug occurs.

I would argue we fix that by pressing ahead with #1121 so that navigating into a PDU has to happen from the homepage and we remove the switcher entirely. We can then remove `selected_audit_year` and `pz_code` from the session entirely.

### DIFF
--- a/project/npda/context_processors.py
+++ b/project/npda/context_processors.py
@@ -19,7 +19,7 @@ def current_audit_period_slug(request):
     if request.resolver_match:
         audit_period_slug = request.resolver_match.kwargs.get("audit_period", None)
 
-     # Temporary hack until all pages migrated over to new URL structure
+    # Temporary hack until all pages migrated over to new URL structure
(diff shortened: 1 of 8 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 59 seconds ago) Please check your changes!

PR #1189, opened 2025-08-02
Fix changing employer during csv upload

### LINKED ISSUE #1186: Switching organisation is broken on the CSV upload page
`submission_employer_selector` doesn't work, so if you try to upload a CSV and are in multiple units you're not presented with the option to switch.

Apologies, I broke this in #1007. I made the `upload_csv` route async and that meant the existing code started to throw a synchronous only exception. For some strange reason, probably due to lazy evaluation of query sets, this error is swallowed within the templates because we do

```
{% if employers|length > 1 %}
```

If you try and render it directly you get the crash

```
{{employers}}
```

I started trying to fix this but since we're thinking about moving navigating between units to the homepage (#1121) then I think probably we just wait for that and change the CSV upload page to have a link to there to do the switching?

mbarton:
Fixes #1186 

Screenshot 2025-08-02 172920

As we can link directly to a different PDU now we can fix this selector without an additional endpoint.

### DIFF
--- a/project/npda/templates/partials/submission_employer_selector.html
+++ b/project/npda/templates/partials/submission_employer_selector.html
@@ -1,4 +1,5 @@
 {% csrf_token %}
+{% load npda_tags %}
 <div role="alert" class="alert border-t-1 border-blue-400 shadow-lg mt-5">
   <i class="fa-2xl fa-info-circle fa-solid"></i>
   <div>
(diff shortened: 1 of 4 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 4 seconds ago) Please check your changes!

Issue #1201, opened 2025-08-18
banner for workforce survey


AmaniKrayemRCPCH:
Heya, 

Are we able to add the following message to the npda platform between 1st September and 12th December. 

"The NPDA 2024/25 Workforce Spotlight Audit is now open. Please submit your response [here](https://eu.surveymonkey.com/r/NPDAWorkforceSpotlight2025). More information is available on the [NPDA spotlights webpage](https://www.rcpch.ac.uk/resources/npda-spotlight-audit-reports)."

mbarton:
@AmaniKrayemRCPCH do you want this on every page?

AmaniKrayemRCPCH:
Every page is fine, if they're able to dismiss it?

If not, then just on log in/homepage.

mbarton:
Once #1214 is merged, I will add this banner using the Django admin to target the login page and homepage:

URL matcher: `\/account\/login.+|\/home`

```html
The NPDA 2024/25 Workforce Spotlight Audit is now open. Please submit your response <a class="underline" href="https://eu.surveymonkey.com/r/NPDAWorkforceSpotlight2025">here</a>. More information is available on the <a class="underline" href="https://www.rcpch.ac.uk/resources/npda-spotlight-audit-reports">NPDA spotlights webpage.</a>
```

I'll then toggle it to enabled on Monday morning.

mbarton:
Banner is live. Leaving this issue open to remind me to turn it off on December 12th

mbarton:
Banner removed

Issue #974, opened 2025-04-25
Process each patient in a separate Celery task


mbarton:
Now we support processing CSV uploads using Celery we think it would be more maintainable to process the rows for each patient as a separate task. We could then remove all the async code that currently keeps uploads fast by calling the validation APIs in parallel. Instead we would rely on having enough workers to process the per patient tasks in parallel.

Issue #562, opened 2025-02-07
Banners


mbarton:
We want to add messaging indicating people need to use the old platform until the start of the next audit year.

Port across the banners implementation from E12: https://github.com/rcpch/rcpch-audit-engine/pull/1076

mbarton:
Branch for this: [mbarton/banners](https://github.com/rcpch/national-paediatric-diabetes-audit/tree/mbarton/banners)

We hit a problem on async routes however since the code uses synchronous only functions like `cache`. There's not really a generic way of fixing this that I can think of so I suggest we look at this again once we've migrated all the async routes to Celery tasks

PR #1213, opened 2025-08-29
Make async route sync

mbarton:
This was left over from migrating to Celery, where the functionality could be called directly from the route (and therefore had to be async compatible) or from the Celery task (sync).

The csv_upload function remains async as it runs external validators (eg dGC API calls) in parallel. (https://github.com/rcpch/national-paediatric-diabetes-audit/issues/974 was raised to replace this with a Celery task per patient and use the inherent parallelism of multiple Celery workers rather than async juggling).

By doing this we unblock #562 (and #1201) - all of the routes will be sync so context processors can happily call sync APIs.

### DIFF
--- a/project/npda/general_functions/csv/csv_upload.py
+++ b/project/npda/general_functions/csv/csv_upload.py
@@ -43,21 +43,21 @@ from project.npda.models import (
     VisitActivity
 )
 
-async def create_csv_submission(pdu, audit_period, csv_file_bytes, csv_file_name, submission_active, user=None, ip_address=None, new_dataframe=None):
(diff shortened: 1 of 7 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 39 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 43 seconds ago) Please check your changes!

Issue #1211, opened 2025-08-22
Roll out new navigation without a feature flag


mbarton:
After https://github.com/rcpch/national-paediatric-diabetes-audit/issues/1209 and https://github.com/rcpch/national-paediatric-diabetes-audit/issues/1210 make the new navigation the default.

Keep the feature flag in case we break unknown workflows and need to put the old one back temporarily.

mbarton:
Tasks todo once we're happy everyone is happy and remove the feature flag:

- Remove old switcher including `view_preference` and `audit_year` endpoints
- Remove `pz_code` and `selected_audit_year` from session (#1195)
- https://github.com/rcpch/national-paediatric-diabetes-audit/issues/1206

PR #1218, opened 2025-09-02
Skip selector page for users in only one unit

mbarton:
When users log in:

- Send them to the new unit selector page if they have more than one unit
  - Even without the new breadcrumb nav turned on, it's better UX
- Send them straight to the unit report if they are only in one unit
  - Unless the new nav feature preview is on, the "home" button still takes you to the old homepage

Once the flag is on by default (#1211) then the home button will take you to the same page you are sent to after you log in.

### DIFF
--- a/project/npda/context_processors.py
+++ b/project/npda/context_processors.py
@@ -1,9 +1,8 @@
 import re
-from datetime import datetime
 from django.conf import settings
 from django.core.cache import cache
-from project.npda.models.audit_period import AuditPeriod
 from project.npda.models.banner import Banner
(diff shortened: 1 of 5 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 42 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 54 seconds ago) Please check your changes!

Issue #1195, opened 2025-08-03
Remove all remaining session data


mbarton:
Blocked on #1121 - without the session data we can't generate data URL links link "Unit Report" etc from non data URL pages like the user list

mbarton:
Blocked on https://github.com/rcpch/national-paediatric-diabetes-audit/pull/1121. For as long as we retain "data URL" links in the header that applies to all pages, we'll need the current audit year and PZ code in the session. If we're happy with the new breadcrumb based navigation we can then do this.

PR #1219, opened 2025-09-02
Remove pdu_choices from session

mbarton:
Part of #1195. This does mean a database query to load the PDU choices on every page but that will go once we roll out the new navigation and the switcher is on a separate page.

Doing this for tech debt reasons but has the nice side benefit that users don't need to log out and back in again when they're granted access to a new PDU to select it.

### DIFF
--- a/project/npda/admin.py
+++ b/project/npda/admin.py
@@ -121,8 +121,6 @@ class SessionAdmin(admin.ModelAdmin):
         "session_key",
         "user_id",
         "pz_code",
-        "organisation_choices",
-        "pdu_choices",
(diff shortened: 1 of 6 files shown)


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 7 minutes and 47 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 54 seconds ago) Please check your changes!

PR #1220, opened 2025-09-02
Replace audit_years in session with context

mbarton:
Part of #1195 

We can replace `audit_years` in the session with looking up the audit periods the user can see when building the context.

Once we roll out the new navigation and remove the global filters switcher we can then remove it from the context entirely.

`selected_audit_year` and `pz_code` have to remain in the session until we do that, otherwise we won't know how to set the selected PDU and audit year in the switcher.


mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech), [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 7 minutes and 54 seconds ago) Please check your changes!

PR #1223, opened 2025-09-03
Fix following direct links whilst logged out

### LINKED ISSUE #1222: Some direct links don't work for logged out users
**With DEBUG=false**

As a logged out user, visit `https://npda.localhost/period/2025-2026/pdu/PZ999/dashboard` (protected by `login_and_otp_required`). After you log in you end up at the two factor setup complete screen.

If you do the same with a class based view route like `https://npda.localhost/period/2025-2026/pdu/PZ999/patients` you are redirected to the page you originally tried to visit.

In the latter case, the `?next=` parameter is passed correctly. In the former it is automatically set to the two factor setup endpoint (`?next=/account/two_factor/setup/).

To compound the confusion:

**With DEBUG=true**

The same issue occurs but then it doesn't matter anyway since you're always sent to the homepage after login.

I discovered this issue because a test started failing on #1221 (https://github.com/rcpch/national-paediatric-diabetes-audit/actions/runs/17407876134/job/49417093352?pr=1221). The test checked that all dashboard routes are inaccessible to logged out users. I added a new route that wasn't class based and the test failed because it ended up with the wrong `next` parameter. A good example of "testing in depth" maybe? I'm not sure, but lets fix anyway. Not reliably working for direct links after logging in is one of my pet hates for web apps, this bug has crept in to literally every system I've ever worked on! So let's also add some unit tests to try and stop it regressing again.

mbarton:
Fixes #1222 

Fix following direct links whilst logged out for views protected with `login_and_otp_required`. Class based views already worked.

Also fixes:

- The password expiry check was bypassed if you followed a direct link to a page
- Our `DEBUG` two factor bypass didn't redirect you to the page once logged in


project/npda/views/decorators.py
```diff
@@ -56,9 +54,7 @@ def sync_login_and_otp_required(request, *args, **kwargs):
             else:
                 return redirect("two_factor:setup")
 
-        login_required(view)
-
-        return sync_login_and_otp_required
+        return login_required(sync_login_and_otp_required)
```
mbarton, line 57: This is the fix for function based views - we need to use `login_required` as we would a decorator, making sure it can return a response including `next`.

Previously the result of this check was being ignored, so it was then up to `check_otp` to fail for users who hadn't logged in yet and therefore sent them to two factor registration.

project/npda/views/npda_users.py
```diff
@@ -722,68 +722,53 @@ def post(self, *args, **kwargs):
...
-    def done(self, form_list, **kwargs):
-        # this will not be called if debug=True
-        response = super().done(form_list)
-        response_url = getattr(response, "url")
-
-        # redirect to home page
-        login_redirect_url = reverse(settings.LOGIN_REDIRECT_URL)
-
-        # Successful 2FA and login
-        if response_url == login_redirect_url:
```
mbarton, line 754: This is why you'd bypass the password expiry check if you logged in following a direct link. I don't really know why this was here, it appears to have been in the original 2fa bypass PR from E12 (https://github.com/rcpch/rcpch-audit-engine/pull/574)

mbarton:
Seen on [STAGING](https://npda-staging.rcpch.tech) (merged by @mbarton 8 minutes and 6 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://npda.rcpch.ac.uk) (merged by @mbarton 8 minutes and 22 seconds ago) Please check your changes!

PR #1227, opened 2025-09-08
Enable new navigation by default

### LINKED ISSUE #1211: Roll out new navigation without a feature flag

mbarton:
Fixes #1211 

Retains the flag in case people want to revert back to the old navigation. Send users to either the new PDU switcher after they log in or if they are only in one PDU send them direct to their unit report.


mbarton:

PR #1259, opened 2025-10-11
Enhance KPI tests and ineligibility tooltips

### LINKED ISSUE #1257: Patient report requires smoking cessation date for non smokers
Patients who are non-smokers are still marked as incomplete in the patient report if they don't have a smoking cessation referral date

eatyourpeas:
### Overview
Fix #1257
Non smokers currently fail if they have not been referred for cessation support, but of course it is only the smokers that should get that.

This PR updates the KPI calculation to exclude non-smokers from the calculation around referrals, and updates the patient report to reflect nonsmokers as ineligible for this measure with a suitable tooltip to reflect this.
A second issue fixed along the way was to add an AuditPeriod fixture to the test suite so that KPI tests can be run individually, as currently AuditProgress is added only at the beginning of all KPI tests.

### Code changes
`kpis.pi` - exclude nonsmokers from cessation referrals
`confttest.py` - create an AuditProgress fixture
`test_patient_report.py` - fix the tests and import the new fixture
`patient_report.py` - alters the annotation for  smoking cessation referrals to return a Char rather than a Boolean to allow different ineligibility types. the template is then updated as well as the ineligiblility tooltips

Updates the KPI 36 documentation to reflect nonsmokers excluded

### Related Issues
closes #1257

### DIFF
--- a/documentation/docs/developer/kpis/additional_processes.md
+++ b/documentation/docs/developer/kpis/additional_processes.md
@@ -42,7 +42,7 @@
 
 **Numerator**: Number of eligible patients with an entry for Date of Smoking Cessation Referral (item 41) within the audit period
 
-**Denominator**: Number of patients with Type 1 diabetes aged 12+ with a complete year of care in the audit period (measure 6)
+**Denominator**: Number of patients with Type 1 diabetes aged 12+ with a complete year of care in the audit period (measure 6) who screen as smokers
 
 **Data Items**: 41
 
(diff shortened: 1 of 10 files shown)


eatyourpeas:
In the process of looking at this I have found there is a test that expects children under 12 asked if they smoke  in the patient report pass anyway which feels to me not right so I have reimplemented the same as smoking cessation - those that are under 12 become ineligible in the patient report table and the test now expects `None` to be returned instead of `True`.

PR #1266, opened 2025-10-22
More testing PDUs

mbarton:
PZ998 external testing can be used by NPDA network managers and pen testers to use the platform without having access to the RCPCH staff email addresses you would see if you used PZ999. Request from NPDA team.

PZ997 I have added as a placeholder for future automated testing

### DIFF
--- a/project/npda/management/commands/seed_functions/paediatric_diabetes_units_seeder.py
+++ b/project/npda/management/commands/seed_functions/paediatric_diabetes_units_seeder.py
@@ -9,6 +9,24 @@ from project.npda.general_functions import fetch_organisation_by_ods_code
 
 logger = logging.getLogger(__name__)
 
+RCPCH_TESTING_PDUS = {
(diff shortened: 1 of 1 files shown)


project/npda/management/commands/seed_functions/paediatric_diabetes_units_seeder.py
```diff
@@ -28,6 +46,11 @@ def paediatric_diabetes_units_seeder():
     pdus = get_all_pz_codes_with_their_trust_and_primary_organisation()
 
     for pdu in pdus:
+        pz_code = pdu.get("pz_code")
+
+        if pz_code in RCPCH_TESTING_PDUS:
```
mbarton, line 51: PZ999 is in the list that comes back from our NHS organisations API but I would argue testing PDUs are polluting that list if we wanted to use it for other things. So I've added these new ones just in the NPDA app itself.

PR #1280, opened 2025-11-26
Export questionnaire data as CSV

### LINKED ISSUE #1020: Enable units who submit data via patient questionnaire to extract the data via CSV
I've had a few emails asking this question and I believe it would be beneficial for users to have this functionality

mbarton:
Closes  #1020 

Replaces the "Raw Data" download button on the submission screen with an "Export as CSV" button for submissions done via questionnaire.

### DIFF
--- a/project/npda/general_functions/csv/csv_download.py
+++ b/project/npda/general_functions/csv/csv_download.py
@@ -1,13 +1,16 @@
 import json
+import csv
+from io import StringIO
 
 from django.apps import apps
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
 
 from ..write_errors_to_xlsx import write_errors_to_xlsx
+from ....constants.csv_headings import CSV_HEADING_OBJECTS, UNIQUE_IDENTIFIER_JERSEY, UNIQUE_IDENTIFIER_ENGLAND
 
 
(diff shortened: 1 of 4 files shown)


project/npda/general_functions/csv/csv_download.py
```diff
@@ -41,3 +45,51 @@ def download_xlsx(request, submission_id):
...
+
+    if is_jersey:
+        HEADINGS_LIST = UNIQUE_IDENTIFIER_JERSEY + CSV_HEADING_OBJECTS
+    else:
+        HEADINGS_LIST = UNIQUE_IDENTIFIER_ENGLAND + CSV_HEADING_OBJECTS
+
+    header = [row["heading"] for row in HEADINGS_LIST]
+    writer.writerow(header)
+
+    visits = Visit.objects.filter(patient__in=submission.patients.all()).select_related("patient").prefetch_related("patient__paediatric_diabetes_units")
```
mbarton, line 65: I think this should only do two queries. We need `prefetch_related` for the `Transfer` model (via the now confusingly named `paediatric_diabetes_units` relation) as the foreign key to patient is defined on `Transfer` - legacy from before we decided to have multiple patients with the same NHS number (across submissions and transferring between units)

PR #1281, opened 2025-11-27
Fix N+1 query in submissions national view

mbarton:
Screenshot 2025-11-27 123511


mbarton:
Screenshot 2025-11-27 132633

It's still not exactly a fast page to render but a little bit faster now!

PR #1313, opened 2026-01-24
Copy patients from previous audit year if using questionnaire

### LINKED ISSUE #920: Copy patients entered via questionnaire across audit years
At the moment you have to enter all the patients again every year. #919 imports them from the old platform, this issue is about copying patients across audit years within the platform

mbarton:
Closes #920 

User can automatically copy patients from the previous audit year when you select questionnaire for this year. Exclude patients who have died or left service.

Screenshot 2026-01-26 132648

Selecting the copy option uses a new action in the submission list `POST` handler to create the new submission and copy the patients. Otherwise the new submission is created on the first `POST` to the patient create form as before.


project/npda/views/submissions.py
```diff
@@ -302,6 +305,48 @@ def post(self, request, *args, **kwargs):
...
+            previous_audit_period = self.audit_period.previous_audit_period()
+
+            next_submission = Submission.objects.get_submission_for_request(self.pdu, self.audit_period)
+            last_submission = Submission.objects.get_submission_for_request(self.pdu, previous_audit_period)
+
+            if next_submission:
+                raise Error(f"Cannot start questionnaire submission. Active submission already exists for this audit period. audit_period={self.audit_period}, previous_audit_period={previous_audit_period}, pdu={self.pdu.pz_code}")
+
+            if last_submission and last_submission.csv_file_name:
+                raise Error(f"Cannot start questionnaire submission. Previous submission was CSV file. audit_period={self.audit_period}, previous_audit_period={previous_audit_period}, pdu={self.pdu.pz_code}")
```
mbarton: There's no technical reason we couldn't do this so perhaps should be supported?

mbarton:
@AmaniKrayemRCPCH @richardshepherd45 could I get your advice? Would you like people to be able to switch from CSV to questionnaire at the start of the audit year and have patients copied automatically? Or just copy patients for those continuing to use the questionnaire. Thanks!

AmaniKrayemRCPCH:
Is it possible to give them the choice? I think some units will want to carry over their patients and others won't.

PR #1319, opened 2026-01-27
2026-Dataset

### LINKED ISSUE #1243: 2026 dataset update
https://www.rcpch.ac.uk/resources/national-paediatric-diabetes-audit-data-submission#_025-26-dataset---information

New questions for 2026 need adding to the platform for both CSV uploads and questionnaire.

They should only apply if the audit year is 26-27 onwards. It's worth considering how much engineering is involved in properly tying this to audit period definitions rather than if/else statements based on the audit year.

Database column names for new questions should be aliased so they clearly indicate they were introduced in 25-26.

### LINKED ISSUE #1339: New 2026 dataset failing tests in CI due to out of memory.
The dataset branch containing the new 2026 dataset changes currently is failing tests in CI due to lack of memory after merging latest changes from Live in. It was working before that. The tests do increase here as we run them for 2026 and 2021 for those measures affected.

I have tried multiple fixes including:
1. adding xdist parallelisation
2. batching / chunking tests and restarting django between
3. adding a flag for 2026 tests, 2021 tests and all the rest.

None of these unfortunately seem to stop the 137 memory error.

eatyourpeas:
### Overview
2026 Dataset

This PR adds the 2026 dataset ready for 1.4.2026. 

A new method is added to the AuditPeriod model that calculates whether the Visit and Patient tables (and forms) should use the 2026 measures or the 2021 measures.

The 2026 measures and lookup functions have been added as well as the csv from the website and a test csv to match the old dataset. The look up functions with alternatives have been added as the new csv has lots of dagger and other non-latin characters which need stripping but including in the alternatives list.

The AuditPeriod instance now is passed from the view to the VisitForm and this is used in the questionnaire to choose which questions and categories to render, and in the csv upload function as well. This means it should just flip over on the 1/4/2026. If people upload csvs against an older audit year, it will expect the old csv.

The test suite has been updated to accept 2021 and 2026 as params, as well as the audit period instance, so in effect all the csv upload tests are duplicated, running once for 2021, and again for 2026. All the new fields should be tested, but there may be more coverage still to add.

Please forgive the rambly nature of this PR. There were lots things added then removed and moved. Funadamental questions for me still remain:

1. The questions relating to:
- learning difficulties
- asd/adh diagnosis
- immunotherapy
To me these are one time events per child and therefore make most sense from a data model point of view in the patient model. However, my guess is they will not be easy to find if we put them in the patient form, so possibly it is easier to put everything in the visit form. The only issue here is that the analysts will have to sweep all the visits to find a single instance to pass the measure. If this is acceptable we can leave it this way. An alternative might be to put them in the Patient model but render them in the Visit form. Or render them in the Patient form in a new tab.

Anyway, otherwise this I think is working and all the tests are passing

image

image


eatyourpeas:
image

eatyourpeas:
@mbarton now ready for review - sorry about the mix up.
After all the to and fro this PR, in addition to the above, finally moves 4 of the new fields to the `Patient` model and `PatientForm`
- `immunotherapy_received`
- `immunotherapy_date`
- `adhd_asd_status`
- `learning_difficulties_status`
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
To make the patient form less unmanageable i have introduced some colours and categories. I wonder if we need some signposting so that people know that these items are here, or whether we leave it up to the NPDA team to do this. Very happy to take suggestions.

mbarton:
NPDA meeting 20/02/2026 - consider wording of the immunotherapy question to cover some patients who have taken part in a blinded trial so it may not be known whether they actually received the treatment or not

project/constants/csv_headings.py
```diff
@@ -294,16 +307,568 @@
...
+        "model": "Visit",
+    },
+    {
+        "heading": "Initial Standard bicarbonate at admission (mmol/l)",
+        "model_field": "initial_bicarbonate_admission",
+        "model": "Visit",
+    },
+)
+
+CSV_HEADING_OBJECTS_2026 = (
```
mbarton: Can we avoid duplicating these definitions?

project/constants/csv_headings.py
```diff
@@ -294,16 +307,568 @@
...
+        "heading": "Initial Standard bicarbonate at admission (mmol/l)",
+        "model_field": "blood_gas_bicarbonate",
+        "model": "Visit",
+        "alternative_headings": [],
     },
 )
 
-ALL_DATES = [
+
+def get_csv_heading_objects(dataset_year=None):
```
mbarton: My instinct is to make `dataset_year` a mandatory parameter? Unless it requires rewriting large amounts of code

project/npda/forms/patient_form.py
```diff
@@ -114,17 +126,110 @@ def __init__(self, *args, **kwargs):
...
+        extra_fields = {
+            "date_leaving_service",
+            "reason_leaving_service",
+            "gp_practice_postcode",
+            "unique_reference_number",
+        }
+        keep_fields = set(allowed_fields) | extra_fields
+
+        # Remove fields that are not relevant to this dataset year
+        for fname in list(self.fields.keys()):
```
mbarton, line 154: My instinct is that this is a lot of code in the constructor and would benefit from moving out, potentially even to a static method. What do you think?

project/npda/general_functions/csv/csv_clean.py
```diff
@@ -42,30 +51,91 @@ def clean_csv_measurement(value):
...
-    if not is_numeric_dtype(df["Stated gender"]):
-        df["Stated gender"] = df["Stated gender"].apply(clean_csv_sex)
-    
+
+def csv_clean(df, dataset_year=2021):
+    # Helper normalisation to robustly match noisy incoming headings
+    def _norm(s: str) -> str:
+        if not isinstance(s, str):
+            return ""
+        s = unicodedata.normalize("NFKD", s)
```
mbarton, line 74: I wonder if this would lose data for CSV uploads? Where we generally store invalid values providing they fit the data type enough

mbarton: Is this is response to a particular need or defensive?

project/npda/general_functions/csv/csv_upload.py
```diff
@@ -102,23 +111,72 @@ async def csv_upload(
...
     """
     Processes standardised NPDA csv file and persists results in NPDA tables
     Returns the empty dict if successful, otherwise ValidationErrors indexed by the row they occurred at
     """
     pdu = submission.paediatric_diabetes_unit
 
+    # Infer dataset_year from submission.audit_period
+    try:
+        dataset_year = submission.audit_period.get_dataset_year()
+    except Exception:
```
mbarton, line 135: Surely fail the upload instead?

project/npda/general_functions/csv/csv_upload.py
```diff
@@ -102,23 +111,72 @@ async def csv_upload(
...
+        dataset_year = 2021
+
+    # But the dataframe itself may contain 2026 headings; prefer detecting from the dataframe
+    # if present so tests that pass a dataframe directly don't rely on seeded AuditPeriod years.
+    from project.npda.general_functions.headings import get_field_heading
+
+    try:
```
mbarton, line 142: Yes I think we should discuss what this is defending against? For user uploads it should just use the expected format, regardless of what is present in the file

project/npda/general_functions/headings.py
```diff
@@ -0,0 +1,213 @@
+"""
+Field headings configuration for NPDA dataset across different years.
+Maps model field names to their display headings from the official dataset documents.
+"""
+
+PATIENT_FIELD_HEADINGS_2021 = {
```
mbarton, line 6: Again I wonder if we should replace/merge this with csv_headings.py?

project/npda/tasks.py
```diff
@@ -28,10 +32,18 @@ def upload_csv_task(submission_id):
 
     logger.info(f"This is the submission to process: {submission} [{submission.id}]")
 
+    # Determine the dataset year from the CSV data (use integer year)
+    try:
+        dataset_year = submission.audit_period.get_dataset_year()
+    except Exception:
+        # Fallback to 2021 for safety
```
mbarton, line 39: As above, I don't know what we're guarding against here? At least it logs at `logger.exception` I guess so we'd get an email

project/npda/views/submissions.py
```diff
@@ -12,7 +12,17 @@
 from django.contrib.auth.mixins import PermissionRequiredMixin
```
mbarton, line 17: @copilot what's changed in this file that isn't formatting changes?

project/npda/views/npda_users.py
```diff
@@ -32,12 +32,7 @@
 from project.npda.models.paediatric_diabetes_unit import PaediatricDiabetesUnit
```
mbarton, line 29: @copilot has anything changed in this file apart from formatting?

mbarton (commented):
Some initial comments from my first look. This is going to be a big PR by the nature of it and the formatting changes make some files harder to review.

I think we will need to run through this in person to fully share the knowledge about what's changed. I also wonder if it can be broken down into smaller individual PRs that then make the diff that introduces the 2026 specific stuff easier to manage (e.g. getting an agent to de-duplicate all the various places we have column/model field config).

eatyourpeas:
These comments are all super helpful thanks @mbarton. Perhaps we should spend an hour to pull this to pieces. There are definitely some dead constants the were necessary at the start and weren't by the end.

project/constants/csv_headings.py
eatyourpeas: The difficulty here is that the csv headers are subtly different for 2026 vs previous. Note that they include extra characters often whch have to be escaped and the definitions in several places have subtly changed.  I found it easier just to have a pre and post 2026 approach- would you rather i changed it to fall back to 2021 if not in the list?

project/constants/csv_headings.py
eatyourpeas: i have made this change

project/npda/forms/patient_form.py
eatyourpeas: i have done this

project/npda/general_functions/csv/csv_clean.py
eatyourpeas: This is because of the headers I think in the new csv containing lots of nonunicode characters. It is meant to esacape them. Perhaps we could review this bit together before merging? I can show you the csv that RCPCH have published as their standard but it is more complex than the previous.

project/npda/general_functions/csv/csv_upload.py
eatyourpeas: oh i see. I was thinking that we would default to 2021 but yes you are right we should possibly raise and error to the user here. Perhaps another one to implement together.

project/npda/views/npda_users.py
eatyourpeas: I don't think so. I think in the process of adding and removing committed items i inadvertently rewound the branch to the start so there maybe noise in here as a result.

project/npda/views/submissions.py
eatyourpeas: I think the `dataset_year` was introduced into the file

eatyourpeas:
Small update for @mbarton:
What with all the linting issues and other updates to live, there were some 50-100 merge conflicts. In the process of fixing this and then fixing the broken tests in fact I have completely rationalised all the CSV headers (which I think was the main feedback from your PR reviews). So in essence the changes here have created a single `ALL_HEADERS` object that is used as the canonical standard for all CSV headers and has `alternative_headings` that are accepted where there are typos/variations/extra non-ASCII characters and so on. This object also contains the `dtype` for parsing, the model and the dataset year to which it applies, so deduplicating all those *_2021 and *_2026 lists. There are custom functions now that pull out all the visits, or all the dates for a give year directly from here so we should hopefully now be in a better place should we come to add new datasets in future. Only one place to update it all.

The tests all pass and I have tested the UI and it toggles between both datasets depending on the audit year selected in the picker.

The linter now is running on all pushes and there is linting automatically added on save to the .vscode_settings.json here as well.

I am reasonably confident it is ready to merge but obvs will take my lead from you.

Happy Easter 🐣

## 2025/2026

image

## 2026/2027

image

eatyourpeas:
Before merging just to say there are several new commits which don't touch any of the implemented functionality, but change the UI work flow in the light of the new dataset, mostly around signposting and messaging to users.

## Guard against uploading a csv with the wrong headers

This is self-explanatory. There are now tests for this too.

image

## User interface items

The main change is to add a dropdown for Audit Period within the PDU list:

image
image

This shows for users who have access to many PDUs or just single access. It also explains to readers who have no permissions to upload to new audit years that they cannot:

image

And by the same token the same approach appears in the PDU dashboard:

image

Here there is signposting to users that are familiar with the interface to remind them that they are viewing the default audit period (whose end date has passed) rather than the new one. There are two ways for those with permissions to click through to the upload screen. 

And equally for submissions that are now old there is messaging to say so.

image

Currently there is no clear decorator that prevents uploading to closed submissions (see #1374) but this PR adds some UI guards to prevent people uploading to old submissions. In particular the messaging in the PDU dashboard tells users that CSVs are still accepted for the previous audit period only while we are in the first quarter of the new audit year. Once the first quarter passes the previous audit period will appear closed in the UI. Checking with @AmaniKrayemRCPCH and @mbarton that they are happy with this behaviour.

I don't think this last point should prevent merging, but do please let me know or add your comments to #1374 for what would be better.

PR #1357, opened 2026-03-25
age distributions calculated from 1st day of audit period

### LINKED ISSUE #1334: Age table - age at start of audit year.
Image

Hello! Can this table within the dashboard show 'Age at the start of audit year', as opposed to their current age. (Apologies if that's what it's already doing). 

That's what we use in the NPDA reports, so it would be best if they were aligned. 

Thanks!

eatyourpeas:
### Overview

fixes #1334

The dashboard has a table where patients in the submission are stratified first by diabetes type, and then by age. The age though is calculated as the patient age at the time the page is requested, unless that falls after the end of the audit period (where it is set at the last day of the audit period)

This changes the calculation to always set the age of the child to their age at the start date of the audit period 

Includes unit tests:

| Test | What it verifies |
| -- | -- |
| `test_patient_ages_no_submission` | Zero counts when no active submission exists |
| `test_patient_ages_age_band_assignment[age_X_in_Y]` (6 parametrized cases) | Each of the 6 age bands (0–2, 2–5, 5–12, 12–16, 16–19, 19–25) gets the right patient |
| `test_patient_ages_calculated_from_audit_start_not_today` | Age is computed from audit_period.start_date, not today — the key invariant of this branch | 
| `test_patient_ages_under_and_over_twelve` | The under_twelve / over_twelve aggregate tallies are correct |
| `test_patient_ages_sex_counts` | Male/female sex buckets are populated correctly |
| `test_patient_ages_filter_by_diabetes_type` | POSTing a specific diabetes type filters the counts |
| `test_patient_ages_all_diabetes_types_selected` | POSTing diabetes_type=0 ("All") returns the full unfiltered count |


mbarton (commented):
Looks good, thanks for fixing! Worth a unit test?

PR #1359, opened 2026-03-26
fix/int-parsing-csv-retinal-results

### LINKED ISSUE #1358: Retinal screening result value error
Reported by @FrancescaAnnan 

```
Select a valid choice. 1.0 is not one of the available choices.; Missing item. Retinal Screening Result and the associated date must all be completed.
```

When I am checking so far there is a date entered for retinal screening and the result is entered as a 1 not 1.0 for all but 3 of the patients, which have a date and no result.

Haven’t changed this column since previous uploads

eatyourpeas:
# Fix: Retinal Screening Result (and all nullable integer fields) silently failing on CSV upload

Closes #1358

## Root cause

Three separate bugs in the pandas 2.x dtype handling chain combine to produce the UI error "Select a valid choice. 1.0 is not one of the available choices." for integer-choice fields (Retinal Screening Result, Treatment, HbA1c Format, etc.) when those columns contain any blank cells in the uploaded CSV.

**Bug 1 —** `csv_parse`: None in list comprehension collapses series to `float64`
When pandas reads a column that has integers mixed with blank cells, it stores it as `float64` (e.g. 1 → 1.0, blank → `NaN`). The original code used .astype('Int64') directly on this, which pandas 2 refuses:

The attempted fix used a list comprehension with `None` as the sentinel for `NA` values, then called `pd.Series([1, None, 2], ...)`. In pandas 2.3.x None is treated identically to NaN during series construction, so the result is still inferred as `float64` — the same problem.

**Fix:** Use pd.NA (pandas' true nullable sentinel) with dtype=object explicit so the series stays object-typed, then .astype('Int64') converts cleanly.

**Bug 2 —** `csv_upload`: `iterrows()` degrades Int64 back to float64 per-row
Even after `csv_parse` successfully produces an Int64 column, `iterrows()` in `csv_upload.py` silently downcasts nullable integer values back to `numpy.float64` when constructing each mixed-dtype row Series (a known pandas behaviour). So the correctly stored integer 1 comes back as `np.float64(1.0)`. The existing `value.item()` call then produces Python float(1.0).

**Fix:** In `csv_value_to_model_value()`, after `.item()`, if the result is an integer-valued float and the model field is any integer type (`isinstance(field, django_models.IntegerField)` — which `PositiveSmallIntegerField` inherits from), cast to int.

**Bug 3 —** Django form: `str(1.0) != str(1)`
Django's `TypedChoiceField` stringifies submitted values before comparing against choices. str(1.0) → "1.0" does not match choice key str(1) → "1", so every integer-choice field with a blank row somewhere else in the column would raise a validation error for all users — silently rejecting correctly entered data.

This is the user-visible symptom that was originally reported.

Non-integer bad data (e.g. 99.5) is still correctly rejected
The list comprehension only converts floats where float(v).is_integer() is True (i.e. 1.0, 2.0). A genuine non-integer float like 99.5 is passed through unchanged, causes .astype('Int64') to raise, and the column is added to `parse_type_error_columns` — so downstream validation flags it as an error exactly as before.

## Changes

| File | Change |
| -- | -- |
| `csv_parse.py` | Use `pd.NA` + `dtype=object` in list comprehension; exclude nullable int columns from the post-cast .where(..., None) which was upcasting Int64 → object |
| `csv_upload.py` |	`csv_value_to_model_value()` casts integer-valued floats to `int` for integer model fields; `import django.db.models` at module level |
| `tasks.py` | Remove debug print statement left from investigation |
| `test_csv_upload.py` | Add `test_decs_float_value_cast_to_int` regression test: asserts that a CSV value of 1.0 in the Retinal Screening Result column is parsed to Int64, survives the upload pipeline, and saves as integer 1 with no validation errors |


eatyourpeas:
As an aside it might be worth having a conversation about this before merging in. Although the tests are all passing, I am not sure any of the csv tests we have are truly integration tests, and we are using a dummy csv that has no gaps for example which is possibly one reason why this is only coming to light now. I have uploaded Fran's csv locally and the error is gone but the numbers of parse errors seems to have gone down also (probably a good thing) so it does look different. It might be good to confirm with a few csvs to be sure we are happy.

mbarton:
Thanks @eatyourpeas. This PR is very difficult to review because of the amount of unrelated linting changes.

If we are going to lint on save we must format the entire codebase first so that diffs are then readable. What is your formatting setup? I am happy to do the no-op PR and then add a PR check to enforce formatting. We either have it or we don't and whilst my preference is not to have it I don't hold that strongly and happy to meet you where you are in your setup.

Agree the tests are a mix of lots of different types at the moment. There are some tests that read the dummy CSV, modify it, write it back and then upload it (using `modify_raw_csv` as your new test does) and some that just modify the dataframe. I'm happy to let an LLM make them consistent, probably all as e2e tests?

I do think that Pandas has proven unsuitable for CSV parsing pipelines. It looks like there's a [nullable integer type](https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html) but it's experimental in v3 with no stability guarantees so probably not right to use here even if it did fix this issue.

Definitely too much work to migrate an existing project but I think [Polars has the proper distinction](https://docs.pola.rs/user-guide/concepts/data-types-and-structures/) between `null` and `NaN`. Worth looking at when we're parsing CSV files in the future.

mbarton (approved):
Thanks for the linting PR, this PR is now much easier to digest!

Copilot and I have fixed conflicts so I'm going to send this out now

PR #1389, opened 2026-04-10
fix: date_leaving_service dropped from Transfer when set on a later CSV row

### LINKED ISSUE #1378: Issue with Patient Data - PZ016
We received the below email. When I checked their data, the patient transfer date doesn't appear to be on the individual patient information tab, so I'm unsure if that's affecting the other aspects of what they have emailed in with? I've checked the CSV submission and there are dates for the patient transfer, so I'm unsure why they're not coming through onto the platform.

Hi

I would be grateful if you could you have a look into these queries I have found after submitting the audit:

Thyroid screen -    92% (65/71), I have counted only 2 patients not complete.  On the report there are 5 patients which have transferred to adults, which I think are still being counted in this figure as not complete.  The report does not show the symbol (incomplete year of care) at the side of these, but these figures then do not add up to total 71 ?

Urine screen -  I have only 4 patients not completed, but report says 80%  35/44   and the report is showing 6 patients who have transferred which I think included ?

Feet – I have only 2 patients not completed, but shows 84%  37/44  -  5 patients who have transferred.

eatyourpeas:
### Problem

When a patient has multiple rows in an uploaded CSV and only a later row carries both `Date of leaving service` and `Reason for leaving service`, the date was silently lost — only the reason was saved to the `Transfer` instance.

This was confirmed against a real submission (PZ016 Q4 2025/26) where 6 patients had a visit row with no transfer data followed by a later row with both fields populated.

### Root cause

`merge_rows_for_patient` iterates `ALL_HEADINGS` and calls `merge_patient_rows_for_column` for each Patient/Transfer field. The `match` statement had a case for `reason_leaving_service` — `smallest_code_with_attached_date`, which correctly filters to rows that carry a date and picks the smallest reason code — but **no case for `date_leaving_service`**.

This meant `rows.iloc[0]` (the first row, used as the patient row downstream) still held `NaT` for the date after the merge. `PatientForm.clean()` then saw reason supplied with no date, added an `"invalid"`-code `ValidationError` on `date_leaving_service`, and `get_valid_transfer_fields` nulled it out — leaving the Transfer with only the reason.

### Fix

**`project/npda/general_functions/csv/csv_merge.py`** — add the missing `case "date_leaving_service"` using `smallest(rows, heading)`, picking the earliest non-null date across all rows for the patient.

**`project/constants/csv_headings.py`** — swap `reason_leaving_service` before `date_leaving_service` in `ALL_HEADINGS`. This controls merge iteration order and is required because `smallest_code_with_attached_date` reads `"Date of leaving service"` via `dropna` to identify eligible rows. If the date case ran first (writing a resolved scalar back to all rows), `dropna` would find no nulls and could select the wrong reason code in the conflicting-values scenario.

### Tests added (`project/npda/tests/test_csv_upload.py`)

| Test | What it covers |
|------|---------------|
| `test_transfer_date_leaving_service_resolved_from_later_row` | Core regression: date+reason only on last of 4 rows, both must save to Transfer |
| `test_transfer_date_and_reason_leaving_service_are_saved` | Both fields set as `pd.Timestamp` directly; verifies the csv_upload pipeline (2021 & 2026) |
| `test_transfer_date_and_reason_leaving_service_saved_via_full_parse` | Full pipeline including `csv_parse` string→Timestamp conversion (2021 & 2026) |
| `test_reason_leaving_service_without_date_is_silently_dropped` | Reason with no date is discarded at the merge step — not an error |
| `test_date_leaving_service_without_reason_raises_error` | Date with no reason raises a validation error on `reason_leaving_service` via `PatientForm.clean()` |

### Documentation

Added two new sections to `AGENTS.md` covering the full CSV pipeline — `csv_parse`/`csv_headings` structure and the `csv_upload` flow including Transfer field handling — so future agents understand the multi-step parse→merge→validate→save path before touching this code.

closes #1378

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -215,3 +215,134 @@ When adding a field that is dataset-year-specific:
 ## Notes on Jersey (PZ248)
 
 Patients at the Jersey PDU (`pz_code == "PZ248"`) use `unique_reference_number` as their identifier instead of `nhs_number`. The helper `_patient_identifier_field(pdu)` in `queries.py` handles this. Tests that check `patient_identifier` should be aware of this distinction.
+
+---
+
+## CSV pipeline — `csv_parse` and `csv_headings`
+
+### Key files
+
+| File | Purpose |
+|------|---------|
+| `project/constants/csv_headings.py` | `ALL_HEADINGS` master list + helper functions |
+| `project/npda/general_functions/csv/csv_parse.py` | Reads a CSV file into a pandas DataFrame, normalises headings, detects the unique identifier, and records parse errors |
+| `project/npda/general_functions/csv/csv_clean.py` | Converts date strings to `pd.Timestamp`, cleans sex/ethnicity/measurement columns |
+
+### `ALL_HEADINGS` structure
+
+Each entry in `ALL_HEADINGS` is a dict with:
+
+```python
+{
+    "heading": "<canonical CSV column name>",
+    "model_field": "<Django model field name>",
+    "model": "Patient" | "Visit" | "Transfer",  # absent for PDU Number
+    "data_type": "string" | "date" | "int64" | "float64",
+    "dataset_years": [2021] | [2026] | [2021, 2026],
+    "alternative_headings": [...],  # optional
+}
+```
+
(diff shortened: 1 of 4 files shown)


mbarton (approved):
Thanks @eatyourpeas!

PR #1392, opened 2026-04-13
Fix adding visits

mbarton:
We were not passing "pdu" in the context (instead "paediatric_diabetes_unit") so the UI was showing that the PDU was inactive


mbarton:
Did a quick bisect with Claude and found the issue arrived in https://github.com/rcpch/national-paediatric-diabetes-audit/pull/1319/changes/c4cb97d999251dae33cb9e860e9307a398239b10 where we picked "paediatric_diabetes_unit" instead of "pdu" in the merge.

One more reason to try and avoid long running feature branches!

PR #1340, opened 2026-03-07
Feature/patient-report

eatyourpeas:
## Summary
This work was largely done by GPT-5.2-Codex using prompt in `patient_report.md`
Essentially looks the same as the old patient report but of course under the hood the queries are all new, depending on a user-level feature flag.

- Added per-user persistent feature flag to switch between KPI and helper queries for patient report.
- Implemented helper query module covering all patient report categories with audit-period scoping and transfer/incomplete-year logic.
- Added care-at-diagnosis countdown/overdue display for 14/90 day thresholds with tests.
- Added retinal screening tooltip showing most recent completion date and parity tests running view suite with flag on/off.
- Fixed copy identifier for incomplete year of care rows and aligned tooltips.
- Better signposting for incomplete measures in tooltips
- Note retinopathy now checks for date in previous and current audit period and posts most recent date
- Most recent dates for treatment now used for measures

## Tests
- docker exec -it national-paediatric-diabetes-audit-django-1 pytest project/npda/tests/view_tests/test_patient_report.py
- runs test for new queries (12 passing)
- runs existing patient report tests twice (once with the flag, once without) [46 passing]

image

image

## Remaining work
- Add django-filter facets in column headers. (maybe as a separate PR?)
- PDF export for patient report (#1238).
- 2026 dataset changes (smoke/vape wording; psych additional support column).
 - Admissions extra fields: initial pH and bicarbonate.
 - Conditional treatment column headings by audit year.

Issue #1272, opened 2025-10-31
Once we have multiple years, flag retinal screening as incomplete if not done in the current year or previous year


mbarton:
Follow up to https://github.com/rcpch/national-paediatric-diabetes-audit/issues/1271

mbarton:
Implemented in #1340

mbarton:
Reopening as I'm going to remove the implementation of this from #1351 for two reasons:

- Address regression #1350 by returning blank if no retinal screen present for over 12s
- The current implementation assumes the patient will have the same primary key in subsequent submissions
  - That's not how the data model works since we decided to not normalise by NHS number

PR #1401, opened 2026-04-14
Fix thyroid to apply to all t1 patients with complete year of care

### LINKED ISSUE #1399: Thyroid update - User Feedback/Issues
Hi, below are a series of queries about the Thyroid update:

PZ122

Good Afternoon, 

I have uploaded an updated CSV yesterday and it changed our Thyroid percentage to 85% from 84% which was correct. Everything else was the same as it was from the CSV uploaded on 02/04/2026. 

However when I logged on today the thyroid percentage had dropped to 73% and a few other things such as CGM’s and pump shad changed even though they were not changed on the CSV file.

I have uploaded the original CSV again today and it is still showing as 73% which is a significant difference to what was showing yesterday with he same CSV file. 

Do you know why this data could have dropped off? Could this please be looked into? 

PZ007

Hi Richard,

Apologies for sending you too many emails, but as you are aware we don’t have many days left for submission, so I am bit stressed, I hope you understand.

A-	I think patient and unit report were aligned before as well, even if they were not, there’s still a difference, on platform home page it says 38 patients missing Thyroid screening and when I run the report to check who are missing then it comes up to 32 patients. Please explain what is the reason or am I missing something?
B-	Please could you clarify the age group you are looking for Thyroid screening? Is it up to 16-17, 17-18 or 18- 19 included, and are you counting those patients who have left the service? Or moved out of area? Please don’t hesitate to contact me if this is not clear.

PZ092

Good morning,

Can I ask you to confirm how our thyroid % are worked out please? It states we have 89%, yet we have only 3 patients whom haven’t been screened but your report says 9?

mbarton:
#1340 and #1398 incorrectly calculated thyroid screening so patients diagnosed after 1 year *before* the start of the current audit period were not marked as complete.

This PR adds a test specifically for that case and fixes the problem by applying thyroid screen to all t1 patients with a complete year of care.

This likely resolves #1399 but we need to check the other units not just PZ122

### DIFF
--- a/project/npda/general_functions/patient_report/queries.py
+++ b/project/npda/general_functions/patient_report/queries.py
@@ -172,8 +172,7 @@ def annotate_health_checks(qs, audit_period):
             output_field=BooleanField(),
         ),
         passed_thyroid_screen=Case(
-            When(Q(dx_over_1y=True) & thyroid_exists, then=True),
-            When(Q(dx_over_1y=False), then=None),
+            When(thyroid_exists, then=True),
             default=False,
             output_field=BooleanField(),
         ),
--- a/project/npda/views/patient_report/patient_report.md
+++ b/project/npda/views/patient_report/patient_report.md
@@ -165,16 +165,6 @@ This shows the mean and median HbA1c (both as IFCC and DCCT) of all visit HbA1cs
 |--------|------------|------------|-------|
 | All HbA1c values | `hba1c` + `hba1c_format` + `hba1c_date` | `hba1c` + `hba1c_date` | In 2026 `hba1c_format` is deprecated; values are always mmol/mol. Queries must not rely on `hba1c_format` being non-null for 2026 data. |
 
-## Shortcomings of the current methodology
-
-1. The queries that power the patient report come from the KPI class which are not optimised for generating patient level views, leading to long queries that are slow to run. The KPI class was originally written for PDU level summaries and does not work well at the patient level.
-2. The patient report should only capture patient visits in the audit period selected
-3. Edge cases appear where patients are flagged as having incomplete measures when they are in fact complete. These include:
-   1. thyroid screening: within the Health Check category, this applies to children with more than 1 year of diagnosis. They must have a thryoid function observation date within the audit year to show complete. If they do not they are incomplete. If they are within the first year of care they should be not required.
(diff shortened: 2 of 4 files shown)


project/npda/views/patient_report/patient_report.md
```diff
@@ -165,16 +165,6 @@ This shows the mean and median HbA1c (both as IFCC and DCCT) of all visit HbA1cs
 |--------|------------|------------|-------|
 | All HbA1c values | `hba1c` + `hba1c_format` + `hba1c_date` | `hba1c` + `hba1c_date` | In 2026 `hba1c_format` is deprecated; values are always mmol/mol. Queries must not rely on `hba1c_format` being non-null for 2026 data. |
 
-## Shortcomings of the current methodology
-
-1. The queries that power the patient report come from the KPI class which are not optimised for generating patient level views, leading to long queries that are slow to run. The KPI class was originally written for PDU level summaries and does not work well at the patient level.
-2. The patient report should only capture patient visits in the audit period selected
-3. Edge cases appear where patients are flagged as having incomplete measures when they are in fact complete. These include:
-   1. thyroid screening: within the Health Check category, this applies to children with more than 1 year of diagnosis. They must have a thryoid function observation date within the audit year to show complete. If they do not they are incomplete. If they are within the first year of care they should be not required.
```
mbarton, line 173: Not sure if this contributed to the issue but no longer relevant to the agents anyway since we don't use `CalculateKPIs` anywhere now

project/npda/general_functions/patient_report/queries.py
```diff
@@ -172,8 +172,7 @@ def annotate_health_checks(qs, audit_period):
             output_field=BooleanField(),
         ),
         passed_thyroid_screen=Case(
-            When(Q(dx_over_1y=True) & thyroid_exists, then=True),
-            When(Q(dx_over_1y=False), then=None),
```
mbarton, line 176: `dx_over_1y` is pretty suss - it's defined further up as diagnosis date less than one year before the audit period start date.

I don't think any of the NPDA measures care about this, perhaps it's a misunderstanding of `not required if diabetes duration < 1 year` that was in the patient report spec.

It's also used for retinal screening ("not required" if `dx_over_1y is false). I think this is probably in the spirit of "don't fail retinal screening if not present as it's biannual". But I think it should only be "not required" for patients with an incomplete year of care. For other patients it's required but biannually so the report shouldn't mark them with the angry yellow tick until we have multiple years of data (#1272).

It is also used for care at diagnosis but in practice doesn't matter since we already filter out patients that are diagnosed within the audit year.

PR #1421, opened 2026-05-09
dynamic units for hba1c

### LINKED ISSUE #1420: Accept HbA1c <20
Playing around with the test unit to prepare for a demo to a new unit. I found that if you enter a HbA1c <20, you get an error:

Image

This doesn't make sense since we now automatically determine the format based on HbA1c value, so they can enter a value <20, and we'll assume it's a %.

eatyourpeas:
### Overview

Adds dynamic unit conversion in the questionnaire for HbA1c
Does not yet change the validation step on save

https://github.com/user-attachments/assets/87e9f110-a52c-4064-bb26-77d5cc7aa5a0

closes #1420


eatyourpeas:
This converts values in the HbA1c field back and forth in the browser depending on whether less than or greater than 20. If not None, on submission of the form, the mmol/mol value will always be passed to the view, so the validation in the questionnaire will never trigger, but will continue to trigger as normal for values in the CSV.

If you are happy with this behaviour then the PR is good to go. If you want us to convert values dynamically in the CSV also then we can do that but it does mean mutating the csv before saving.

AmaniKrayemRCPCH:
Looks great! Very happy with this. 

Would prefer not to change the values in the CSV. Happy for those to be left as entered.

PR #1428, opened 2026-05-17
feat: national lead-centre bubble map on submissions page

eatyourpeas:
## Summary

Adds a proportional symbol (bubble) map to the submissions page, visible to audit team admins only. Upgrades the `@rcpch/imd-map` library to v0.5.0 which introduced the `setLeadCentres()` API used here.

### Library upgrade
- `@rcpch/imd-map` 0.4.0 → 0.5.0 (CDN + SRI updated in `imd_map_card.html` and `AGENTS.md`)

### New query — `all_pdus_t1dm_bubble_map_data(audit_period)`
- Added to `queries.py` alongside the existing dashboard functions
- Returns one entry per active PDU that has geocoordinates and T1DM patients in the audit period
- Each entry carries `lat`, `lon`, `label`, `pz_code`, `patient_count` (eligible T1DM), `median_hba1c` (mmol/mol)
- ~4 DB queries regardless of PDU count; reuses same eligibility filters and HbA1c normalisation as `hba1c_stats_by_diabetes_type()`

### Submissions view
- `bubble_map_centres`, `RCPCH_DEPRIVATION_TILES_URL`, `RCPCH_CENSUS_PLATFORM_TOKEN` added to context for audit team members

### Submissions template
- Bubble map rendered below the column chart / stats block (audit team only)
- Bubble **size** = T1DM patient count; bubble **colour** = median HbA1c on a blue→yellow→red gradient
- Legend starts collapsed; scoped CSS makes it scrollable if taller than the map container

## Testing
- All 69 existing patient-report tests pass
- Linter clean

image

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,11 +71,11 @@ All dashboard functions use `build_base_queryset()` as their base and respect `a
 
 The dashboard IMD map in `project/npda/templates/dashboard/map_chart_partial.html` uses the browser bundle from `@rcpch/imd-map`.
 
-- Current bundle version: `0.5.0`
+- Current bundle version: `0.5.1`
 - Script URL:
-  `https://cdn.jsdelivr.net/npm/@rcpch/imd-map@0.5.0/dist/umd/rcpch-imd-map.min.js`
+  `https://cdn.jsdelivr.net/npm/@rcpch/imd-map@0.5.1/dist/umd/rcpch-imd-map.min.js`
 - Current SRI:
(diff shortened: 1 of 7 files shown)


eatyourpeas:
ok i went a bit OTT here but i do think this is much better.
This is only visible to admin so should not affect any users but improves experience for the audit team and tries out the maps.
I am secretly interested to know some of the latest national level data (however imperfect it might be) and this implementation makes the submission list smaller, easier to see and searchable. The maps now show two criteria (patient number in bubble size, another metric by colour) and then there is the submission list as before.

image
image
image
image

PR #1452, opened 2026-06-17
Fix accessing user data by guessing PK

### LINKED ISSUE #1451: Logged in users can access other users details by guessing the ID
As found in the latest pen test:

The application allowed Reader, Editor, and Coordinator privileged accounts to access the information of users other than those on the allowed list. The following are the steps to reproduce the issue:

Step-1:
The tester logged in to the application using reader@test.rcpch.ac.uk user account and accessed the accessible user list, which listed only four accounts, as shown in Figure-1:

Step-2:
The tester clicked on Reader user to edit the information, which redirected the user to https://<redacted_staging_url>/npda_users/113/update URL, as shown in Figure-2:

Step-3:
Further, the tester observed that the URL contained a REST parameter for the user ID value, which was a numerical value and could be easily iterated with incremental or decremental values. Then, the tester altered the user ID value to 114, and the respective user's information was accessible, as shown in Figure-3:

Step-4:
Hence, the tester iterated the user ID value with automation and accessed all possible users' information, as shown in Figure-4:

mbarton:
Fixes #1451 and adds a regression test.

I broke this when we rolled out the data URL changes. `PDUPermissionMixin` had code to do this but was not used for the users view (as it's not PDU specific). I've removed the user case from that view as it's unused.

### DIFF
--- a/project/npda/tests/permissions_tests/test_npda_user_model_actions.py
+++ b/project/npda/tests/permissions_tests/test_npda_user_model_actions.py
@@ -140,6 +140,29 @@ def test_npda_user_list_view_users_can_only_see_users_from_their_pdu(
 
 
 @pytest.mark.django_db
+def test_npda_user_update_view_users_can_only_see_users_from_their_pdu(
+    seed_groups_fixture,
+    seed_users_fixture,
+    seed_audit_periods_fixture,
+    client,
+):
+    ah_user = NPDAUser.objects.filter(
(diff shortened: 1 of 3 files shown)


mbarton:
This has deployed but for some reason prout hasn't updated the labels yet. The prout debug page shows it as deployed and the latest_git_commit embedded in the HTML of the page is correct

PR #1473, opened 2026-07-15
Make patients in transfer show incomplete year of care in patient list

### LINKED ISSUE #1413: Incomplete year of care in patient data includes those who transferred
The filter for incomplete year of care in the patient data view is wrong. It only accounts for diagnosis in the current audit year and not for transfers.

Existing (wrong) query:

https://github.com/rcpch/national-paediatric-diabetes-audit/blob/f525dcbe3ba3ba02139497b815928e07bc51b665/project/npda/views/patient.py#L150

New (correct) query as used in the patient report:

https://github.com/rcpch/national-paediatric-diabetes-audit/blob/f525dcbe3ba3ba02139497b815928e07bc51b665/project/npda/general_functions/patient_report/queries.py#L67

mbarton:
Fixes #1413

### DIFF
--- a/project/npda/tests/view_tests/test_patient_views.py
+++ b/project/npda/tests/view_tests/test_patient_views.py
@@ -1,0 +1,71 @@
+from http import HTTPStatus
+
+import pytest
+from dateutil.relativedelta import relativedelta
+from django.urls import reverse
+
+from project.npda.models.audit_period import AuditPeriod
+from project.npda.models.npda_user import NPDAUser
+from project.npda.models.transfer import Transfer
+from project.npda.tests.factories.patient_factory import PatientFactory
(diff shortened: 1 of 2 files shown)


mbarton:
Test failure as it is calling the NHS organisations API and CI doesn't have credentials. Before patching it out in the tests I'm going to take another look at why we call the API from the patient list in the first place. Ideally we wouldn't have n+1 queries to an external API.

**Update**: ah looks like it's a backfill for the PDU geocordinates so just need to make sure they're available in the test.

**Another update**: I decided to remove the backfill since all PDUs in live have geocoordinates now (https://github.com/rcpch/national-paediatric-diabetes-audit/pull/1475)

PR #1480, opened 2026-07-21
Manage employer security fixes

mbarton:
As discovered through our Fable 5 security review.

**IDOR: unscoped employer management (npdauser_pdu_update)**

`project/npda/views/npda_users.py` — the `npdauser_pdu_update` view is gated by `can_transfer_npda_lead_centre` + `change_npdauser`, but unlike `NPDAUserUpdateView.dispatch` it **never checks that the requesting user shares a PDU with (or administers) the target user** `pk`, and it operates on `OrganisationEmployer` rows by a **POST-supplied `organisation_employer_id` with no ownership scoping**:

- **Delete branch:** `OrganisationEmployer.objects.filter(pk=request.POST.get("organisation_employer_id")).delete()` — deletes *any* employer row by enumerating integer PKs, so a coordinator can remove **any user from any PDU** across the whole instance (account lockout / access-control tampering / DoS).
- **Set-primary branch:** same unscoped `pk`, allowing manipulation of another user's primary-employer state.
- **Add-employer branch:** the *target PDU* is correctly checked against the requester's PDUs, but the *target user* `pk` is not — so an arbitrary existing user can be added to the requester's PDU (granting that user access to its patient data).

### DIFF
--- a/project/npda/tests/permissions_tests/test_npda_user_model_actions.py
+++ b/project/npda/tests/permissions_tests/test_npda_user_model_actions.py
@@ -885,6 +885,49 @@ def test_users_can_download_report(
 
 
 @pytest.mark.django_db
+def test_coordinator_cannot_delete_employer_affiliation_for_other_pdu_user(
+    client,
+    seed_groups_fixture,
+    seed_users_fixture,
+    seed_audit_periods_fixture,
+):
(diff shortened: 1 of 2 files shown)


mbarton:
@copilot can you pls run s/lint —fix on this PR and commit and push the changes. Thanks!

